PR #3, opened 2016-06-13
Use math.fsum instead of sum for floats, and remove extra math import.

ssbr:
(Google is added to COPYING because it prefers that the license file contain the legal copyright owners, and wishes to retain ownership of this patch.  The COPYING change is not a strict requirement, and you or I can revert the COPYING file. [Sorry!])

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,5 +1,5 @@
 The MIT License (MIT)
-Copyright (c) 2016 Victor Stinner.
+Copyright (c) 2016 Victor Stinner and Google Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the
--- a/perf/__init__.py
+++ b/perf/__init__.py
@@ -47,12 +47,10 @@ try:
         # Wrapper to hide the xbar parameter, to be portable with Python 2
         return _stdev(data)
 except ImportError:
-    import math
-
     def mean(seq):
         if not seq:
             raise ValueError("sequence seq must be non-empty")
-        return float(sum(seq)) / len(seq)
+        return float(math.fsum(seq)) / len(seq)
 
     def stdev(seq):
         seq = [float(value) for value in seq]
@@ -61,7 +59,7 @@ except ImportError:
 
         c = mean(seq)
         squares = ((x - c) ** 2 for x in seq)
-        return math.sqrt(sum(squares) / (len(seq) - 1))
+        return math.sqrt(math.fsum(squares) / (len(seq) - 1))
 
 
 _TIMEDELTA_UNITS = ('sec', 'ms', 'us', 'ns')
@@ -552,7 +550,7 @@ def _pooled_sample_variance(sample1, sample2):
     mean2 = mean(sample2)
     squares2 = ((x - mean2) ** 2 for x in sample2)
 
-    return (sum(squares1) + sum(squares2)) / float(deg_freedom)
+    return (math.fsum(squares1) + math.fsum(squares2)) / float(deg_freedom)
 
 
 def _tscore(sample1, sample2):


mahmoud:
+1 for removing Google from the license if this is accepted. This is a minor best practices change, barely more than PEP8. Not appropriate for asserting equal copyrights.

ssbr:
It isn't about what is "appropriate", but what is "legally correct".  Perhaps it would be best to add an AUTHORS file so that everyone's IP contribution, no matter how trivial, is attributed?

Note that as far as I am aware, leaving it out or adding it in does not change whether this contribution is legally owned by Google or not, it just changes whether or not your copyright statement is accurate. (As perceived by some lawyercats I guess.)  If you don't care about accuracy, or you disagree with this interpretation of the law, then it would be best to drop it, but I don't think appropriate-ness is a particularly good reason.

(None of this is legal advice, IANAL, please don't sue my employer, or if you do, please don't drag me into court as a witness! (And don't sue me either!))

mahmoud:
For someone who is not a lawyer, you have a very exacting definition of appropriate. 

I have submitted a lot of PRs to a lot of projects. I don't appear in the COPYING, AUTHORS, or LICENSE. Not every change qualifies as original authorship. Facts are not copyrightable, and Google goes further to say that APIs are not copyrightable. A change as minor as this does not qualify either.

ssbr:
I ninja-edited my comment, sorry. (Maybe reread if you like.)

It is required that all our (Google employees) patches are sent the Google copyright modification with this so as to preserve the correctness of the LICENSE file, but we are also encouraged not to irritate project owners while we're at it, and remove the notice if they don't want it.  (I've failed at the not irritating people part. Sorry. I'm not good at this.)

I realize this is weird and not very normal for open source projects. I don't know why Google is so insistent on this, I am just following the policy.

I don't know and don't want to speculate on whether this actually is not after all Google's copyright. If it is important, I can ask our lawyers what they think.

ssbr:
Deleting comments per lawyercat request.

Please note that this change is optional and not necessary. I am just doing what I am asked to do.

ssbr:
Deleted comments per lawyercat suggestion, in case they are misinterpreted as expressing a legal opinion on behalf of Google, which they are not. I have no actual knowledge of why Google should be added to the copyright declaration.

Instead I will quote somebody:

"Hey, just so you know, you are welcome to use what you want as a notice, but doing that just makes the notice wrong, it doesn't make you actually own anything.  This is true when we are talking about *any* contributions, not just google's.   If you care, great, if you don't, also fine, we are happy to go with the flow.  We just generally try to keep them correct when we contribute unless folks tell us otherwise."

The change is definitely optional and I can't stress that enough.

I am not touching any legal speculation about whether Google owns this or not with a ten foot pole.

mahmoud:
Don't fret, you haven't upset me. Google's aggressive policies in this area are nothing new, and I was all too aware of them.

You *have* made my comments a little silly just floating out there. Sort of like that lawyer quote, it's not clear how they apply here. There's a lot of talk about right and wrong and correct and incorrect. IP is nowhere near this cut and dry. It's unfortunate that Google's lawyers portray it this way. I know some good lawyercats you can speak to if you'd like second opinions.

haypo:
Hi, can you please update your pull request?

When I read it, I saw that the current copyright was not right. Red Hat is the real owner, not myself :-) I'm ok to add Google copyright, I only care of the license: MIT :-)

I saw that statistics is also available on Python 2.7, so I removed perf.mean and perf.stdev, and use directly the statistics module. This module also provides median() ;-)

Your change in _pooled_sample_variance() is still useful.

ssbr:
Done, I think. Sorry, I'm not very good with git/github and I'm not sure if I accidentally added or removed a newline in COPYING.

haypo:
Thanks.

PR #6, opened 2016-07-12
[Core] Refactoring the call command use a dict and functools to dispatch

hobbestigrou:
the action and call the right function.


haypo:
Can you please merge your two changes? It looks like tests are failing with the first change, I would prefer to avoid a temporary failure if it can be avoided ;-)

haypo:
Thanks.

PR #10, opened 2016-07-14
[Documentation] Use pybench command in example.

hobbestigrou:


haypo:
Please document the "python3 -m perf ..." command at the top of the modified page, to explain that it's the same than "pybench" but it allows to control the Python control. It matters a lot when you use the timeit command!

doc/api.rst
```diff
@@ -58,7 +58,7 @@ This command requires the ``scipy`` dependency: see :ref:`Install perf
 
 Example::
 
-    $ python3 -m perf hist_scipy telco.json
+    $ pybench hist_scipy telco.json
```
haypo: Oops, this command of the current doc is wrong. It must be "python3 hist_scipy.py"

doc/cli.rst
```diff
@@ -283,13 +283,13 @@ Example
...
 
-    $ python3 -m perf timeit 1+1
+    $ pybench timeit 1+1
     .........................
     Median +- std dev: 11.7 ns +- 0.1 ns
 
 Use ``-v`` to enable the verbose mode::
 
-    $ python3 -m perf timeit -v 1+1
+    $ pybench timeit -v 1+1
```
haypo: I may add a parameter to specific the Python binary, but until a parameter is added, I prefer to write "python3 -m perf timeit ..." just to be more explicit about the tested Python version.

haypo:
This PR is now outdated.

PR #18, opened 2017-01-07
fix for Windows with pipes

yattom:
This is new PR to replace #17 

By request from @haypo  to use pipes, file handle instead of fd is passed to the child process.  Would you please review again?

Tested on:

- Windows 10 Professional 64bit
   - Python 2.7.11
   - Python 3.5.2
- Ubuntu 14.04.1
   - Python 2.7.6
   - Python 3.4.3

There is a difference in creating a inheritable file handle between Python 2 and 3.  I referred this article https://digitalenginesoftware.com/blog/archives/47-Passing-pipes-to-subprocesses-in-Python-in-Windows.html

I created HandleOfPipe class to wrap the difference.  The class is for Windows only.  It should be possible to also put POSIX pipe into the class, but I thought it's too much.

### DIFF
--- a/perf/_utils.py
+++ b/perf/_utils.py
@@ -365,3 +365,32 @@ else:
     pipe_cloexec = os.pipe
 
     # In Python 3.4, file descriptors are non-inheritable by default (PEP 446)
+
+
+if MS_WINDOWS:
+    import msvcrt
+    if sys.version_info < (3, 0):
+        import _subprocess
+
+    class HandleOfPipe:
+        '''
+        On Windows, fd cannot be inherited but file handles can.
+        Creating inheritable handle differs in Python 2 and 3.
+        This class wraps the difference.
+
+        see also: https://digitalenginesoftware.com/blog/archives/47-Passing-pipes-to-subprocesses-in-Python-in-Windows.html
+        '''
+        def __init__(self, fd):
+            handle = msvcrt.get_osfhandle(fd)
+            if sys.version_info >= (3, 4):
+                os.set_handle_inheritable(handle, True)
+            elif sys.version_info < (3, 0):
+                curproc = _subprocess.GetCurrentProcess()
+                self.duplicated_handle = _subprocess.DuplicateHandle(curproc, handle, curproc, 0, True, _subprocess.DUPLICATE_SAME_ACCESS)
+                handle = int(self.duplicated_handle)
+            self.handle = handle
+
+        def close(self):
+            if sys.version_info < (3, 0):
+                # duplicated handle must be closed in parent process
+                self.duplicated_handle.Close()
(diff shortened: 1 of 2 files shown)


perf/_utils.py
```diff
@@ -365,3 +365,32 @@ def pipe_cloexec():
...
+        This class wraps the difference.
+
+        see also: https://digitalenginesoftware.com/blog/archives/47-Passing-pipes-to-subprocesses-in-Python-in-Windows.html
+        '''
+        def __init__(self, fd):
+            handle = msvcrt.get_osfhandle(fd)
+            if sys.version_info >= (3, 4):
+                os.set_handle_inheritable(handle, True)
+            elif sys.version_info < (3, 0):
+                curproc = _subprocess.GetCurrentProcess()
```
haypo: I don't understand you have to duplicate the handle on Python 2. What is responsible to close this duplicated handle? Why not using something like os.set_handle_inheritable() on the other Python versions? You can probably get SetHandleInformation from ctypes, and hardcode HANDLE_FLAG_INHERIT to 1:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724935(v=vs.85).aspx

perf/_utils.py
```diff
@@ -365,3 +365,32 @@ def pipe_cloexec():
...
 
     # In Python 3.4, file descriptors are non-inheritable by default (PEP 446)
+
+
+if MS_WINDOWS:
+    import msvcrt
+    if sys.version_info < (3, 0):
+        import _subprocess
+
+    class HandleOfPipe:
```
haypo: This object looks very complex just to set a flag on a handle, I don't think that it's needed.

perf/_utils.py
yattom: hmm... I'm not familiar with Win API very well, but MSDN suggests to use CreateFile or DuplicateHandle to make inheritable handles... https://msdn.microsoft.com/en-us/library/windows/desktop/ms683463(v=vs.85).aspx

Still, os.set_handle_inheritable() sure does as you suggested, so I'll try to reproduce it in Python 2.

      DWORD flags = inheritable ? HANDLE_FLAG_INHERIT : 0;
      if (!SetHandleInformation((HANDLE)handle, HANDLE_FLAG_INHERIT, flags)) {
          ...

haypo:
Thank you for your Pull Request, but I wrote my own implementation differently: don't duplicate the handle, but use set_handle_inheritable() on all Python versions => commit 43794b3795250e3b5930e11a4d9c4ee347a021d0

I also updated the test_pipe() of test_runner.py to handle correctly the --pipe parameter.

I also fixed test_timeit which failed on Windows.

I just released perf 0.9.3 which should now work perfectly on Windows. Can you please check?

yattom:
Thank you for your fix.  I'm sorry to be slow and not being much of help.

I checked the new 0.9.3 with Windows OS and found an issue.  I proposed #19.  This time it's a clean patch!

PR #19, opened 2017-01-17
Fix Windows 32bit issue

yattom:
perf 0.9.3 again runs on Windows but it fails with 32bit runtime of Python 2.7 on 64bit platform.

Windows 10 64bit + Python 3.6 32bit -> success
Windows 10 64bit + Python 3.6 64bit -> success
Windows 10 64bit + Python 2.7.11 64bit -> success
Windows 10 64bit + Python 2.7.11 **32**bit -> **fail**

    C:\Python27>Scripts\pip install perf
    ...
    
    C:\Python27>python -m perf timeit --loop 1000 pass
    Error when running timeit benchmark:
    
    Statement:
    'pass'
    
      File "C:\Python27\lib\site-packages\perf\_timeit.py", line 203, in bench_timeit
        runner.bench_sample_func(name, timer.sample_func, **kwargs)
      File "C:\Python27\lib\site-packages\perf\_runner.py", line 541, in bench_sample_func
        return self._main(name, sample_func, inner_loops, metadata)
      File "C:\Python27\lib\site-packages\perf\_runner.py", line 508, in _main
        bench = self._master()
      File "C:\Python27\lib\site-packages\perf\_runner.py", line 768, in _master
        bench = self._spawn_workers()
      File "C:\Python27\lib\site-packages\perf\_runner.py", line 724, in _spawn_workers
        suite = self._spawn_worker(calibrate)
      File "C:\Python27\lib\site-packages\perf\_runner.py", line 644, in _spawn_worker
        warg = wpipe.to_subprocess()
      File "C:\Python27\lib\site-packages\perf\_utils.py", line 432, in to_subprocess
        set_handle_inheritable(self._handle, True)
      File "C:\Python27\lib\site-packages\perf\_utils.py", line 370, in set_handle_inheritable
        ok = SetHandleInformation(handle, HANDLE_FLAG_INHERIT, flags)
    ValueError: Procedure called with not enough arguments (12 bytes missing) or wrong calling convention

This is a calling convention issue.  In _utils.py, ctypes.**cdll**.kernel32 should be ctypes.**windll**.kernel32.  I confirmed it works on Python 2.7 32bit and 64bit.

### DIFF
--- a/perf/_utils.py
+++ b/perf/_utils.py
@@ -362,7 +362,7 @@ if MS_WINDOWS:
         from ctypes import WinError
 
         HANDLE_FLAG_INHERIT = 1
-        SetHandleInformation = ctypes.cdll.kernel32.SetHandleInformation
+        SetHandleInformation = ctypes.windll.kernel32.SetHandleInformation
 
         def set_handle_inheritable(handle, inheritable):
             flags = HANDLE_FLAG_INHERIT if inheritable else 0


haypo:
I didn't know that ctypes.cdll.kernel only works on 64-bit. Does
ctypes.windll.kernel32 work on 32 and 64 bit?

yattom:
>  Does ctypes.windll.kernel32 work on 32 and 64 bit?

Yes, I checked both on Python 32bit and 64bit on 64bit OS.  I don't have 32bit OS at my hand.

yattom:
Actually cdll and windll both access the same dll through different calling convention *cdecl* and *stdcall* (not that I understand it fully... it's just as same as in regular C programming I guess.)

https://docs.python.org/2.7/library/ctypes.html#loading-dynamic-link-libraries

> You load libraries by accessing them as attributes of these objects. cdll loads libraries which export functions using the standard cdecl calling convention, while windll libraries call functions using the stdcall calling convention. oledll also uses the stdcall calling convention, and assumes the functions return a Windows HRESULT error code. The error code is used to automatically raise a WindowsError exception when the function call fails.

https://msdn.microsoft.com/en-us//library/zxk0tw93.aspx

> The __stdcall calling convention is used to call Win32 API functions.

haypo:
Sorry, I forgot your pull request :-( It's now merged into the future perf 1.1 version.

The workaround is to use Python 3.

yattom:
thanks!

PR #30, opened 2017-08-07
Update system show exit code

Lothiraldan:
In case the system is not ready for benchmarking, makes system show exits with
return code 2 so bash scripts could put 'python -m perf system show' directly
without greping for the output.

### DIFF
--- a/perf/_system.py
+++ b/perf/_system.py
@@ -1070,5 +1070,5 @@ class System:
         self.render_messages(action)
         if self.errors:
             sys.exit(1)
-        if self.tuned is False:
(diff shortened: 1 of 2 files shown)


perf/_system.py
```diff
@@ -1068,3 +1070,5 @@ def main(self, action, args):
         self.render_messages(action)
         if self.errors:
             sys.exit(1)
+        if self.tuned is False:
```
haypo: this test is really wrong, please just write "if not self.tuned:"

haypo:
Thanks.

PR #35, opened 2017-09-24
Remove max_time parameter from Runner

### LINKED ISSUE #32: Max time is not used/implemented
Good day.

As far as I can see `max_time` of `Runner` is not used.
First I was going to implement it but I've found out that there is also a commit da7ede7f11ea70a70131dfb15436d2be9358f1e8 which notes attribute as removed.

Are there any plans for that?

Best regards.

gagoman:
Closes #32

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -485,7 +485,7 @@ BenchmarkSuite class
 Runner class
 ------------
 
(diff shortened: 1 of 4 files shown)


gagoman:
Done. Any deprecation policy for the future? E.g. `warnings.warn`

haypo:
> Done

Thanks, I merged your PR.

> Any deprecation policy for the future? E.g. warnings.warn

I don't know. We can discuss that later, if we have something to deprecate :-)

Issue #39, opened 2018-02-15
os.pread not available in Python2


stefantalpalaru:
Error when running "pyperf system show":
```
  File "/usr/lib64/python2.7/site-packages/perf/_system.py", line 146, in read_msr
    data = os.pread(fd, size, reg_num)
AttributeError: 'module' object has no attribute 'pread'
```

vstinner:
Hum, maybe you should use Python 3 :-)

stefantalpalaru:
> Hum, maybe you should use Python 3 :-)

I'm as inclined to use use Python3 instead of Python2 as you are to use Perl6 instead of Perl5.

Seriously, though, I'm packaging your software and it advertises Python2 support which is pretty important for comparing benchmarks between languages and implementations using the "performance" module.

vstinner:
Can you try to write a patch to use seek()+read() when pread() on Python 2?

vstinner:
Fixed by commits e82849af0e59ec792ea2320840f8420e3328422a and 5696ed64bfe69a61f5c69e17656777e0b03f44ff.

PR #40, opened 2018-02-15
enable MSR operations on Python2 - issue #39

stefantalpalaru:


perf/_system.py
```diff
@@ -143,7 +143,8 @@ def read_msr(self, cpu, reg_num, use_warnings=False):
         try:
             fd = os.open(path, os.O_RDONLY)
             try:
-                data = os.pread(fd, size, reg_num)
+                os.lseek(fd, reg_num, os.SEEK_SET)
```
vstinner: I would prefer to use pread() if available:

if hasattr(os, 'pread'):
   ... pread ...
else:
  ... seek ...
  ... read ...

vstinner:
Thanks! I merged your PR.

PR #46, opened 2018-06-03
Add teardown option

### LINKED ISSUE #34: Add a new clearup stage for timeit
I propose to add a clearup stage for timeit. The reason is today when I run perf to benchmark a program, the program will generate a cache file first time and then always use it. I wish to have a separate clear stage, just like the setup stage, to clean it.

gagoman:
Closes #34 

---

I want to add some extra tests and update the doc. Right now PR is to get the feedback if there are something to fix/improve. I have checked if I had made breaking changes. Seems like no.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -548,7 +548,7 @@ Runner class
 
       See the :ref:`bench_func() example <bench_func_example>`.
 
-   .. method:: timeit(name, stmt, setup="pass", inner_loops=None, duplicate=None, metadata=None, globals=None)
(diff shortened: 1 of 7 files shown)


vstinner:
> I have checked if I had made breaking changes.

The tests pass on this PR, so it's fine :-) If you would break something, the problem would be that perf lacks a new test in that case! Don't worry.

gagoman:
I've added docs. Working on `Timer` tests now.

gagoman:
@vstinner which docs are missing?

I've added some tests for the `Timer`. They duplicate code so I'm not sure that they are good.

vstinner:
Oh, I see the documentation. Maybe I missed it at the first read, sorry. I now merged your new parameter, thanks.

PR #52, opened 2019-08-30
Store platform.python_compiler() in metadata

sir-sigurd:


vstinner:
Merged. The Travis CI job is not mandatory, so I was able to merge your PR. I tested locally: "tox -e py3" pass ;-)

PR #53, opened 2019-08-30
Fix Python 3.4 job on .travis.yml

sir-sigurd:


sir-sigurd:
I updated PR, is that what you wanted?

vstinner:
Thanks, merged.

PR #55, opened 2019-10-03
Also include SystemDrive environment variable

zooba:
While running benchmarks against 3.8.0rc1, I noticed a `%SystemDrive%` folder being created in my working directory. Turns out some parts of Windows assume that this environment variable will be set (probably incorrectly, but 🤷‍♂️). There were also some weird crashes that were resolved by providing this variable.

Ensuring it is passed along with `--inherit-environ` is a workaround, but including it by default seems more robust.

### DIFF
--- a/pyperf/_utils.py
+++ b/pyperf/_utils.py
@@ -350,7 +350,7 @@ def abs_executable(python):
 def create_environ(inherit_environ, locale):
     env = {}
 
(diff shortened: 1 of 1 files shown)


vstinner:
Yeah, pyperf is quite strict, but it's to help to get more reproducible benchmarks! Thanks for your contrib, I merged it. Tell me if you need a release.

PR #59, opened 2019-10-14
tests: Require unittest2 on Python 2

jayvdb:
Closes https://github.com/vstinner/pyperf/issues/57

### DIFF
--- a/pyperf/tests/__init__.py
+++ b/pyperf/tests/__init__.py
@@ -14,18 +14,20 @@ try:
 except ImportError:
     import mock   # noqa
 try:
-    # Python 2.7
-    import unittest2 as unittest   # noqa
-except ImportError:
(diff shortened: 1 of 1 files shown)


pyperf/tests/__init__.py
```diff
@@ -27,6 +22,12 @@
 
 from pyperf._utils import popen_communicate, popen_killer
 
+if sys.version_info[0] == 2:
```
vstinner: I would prefer to reuse six.PY2 or six.PY3 for that.

jayvdb: no worries.

I noticed a few other uses of `sys.version_info`, so used that.  Turns out there are only five, and two are replacable with six.PY2/PY3.

vstinner:
Thanks!

PR #67, opened 2020-05-07
Project moved to https://github.com/psf/pyperf

vstinner:


vstinner:
Replaced with PR #68.

PR #77, opened 2020-09-14
performance project was renamed to pyperformance

ichard26:
I had never heard of the `performance` project so I clicked the link out of curiosity. After seeing the 404 it dawned on me that `performance` was probably the former name of `pyperformance`, a project I had seen before. A quick check of the change log confirmed my suspicion :-)


pablogsal (approved):
Thanks for the PR @ichard26 !

PR #84, opened 2020-11-02
Enhance geometric mean formatting

vstinner:


vstinner:
I propose to replace "0.92 (faster)" with "1.09x faster". IMO it's easier to understand. Also, "1.22 (slower)" becomes "1.22x slower".

@methane @pablogsal @serhiy-storchaka: What do you think?

For example, speed.pypy.org says:

> The geometric average of all benchmarks is **0.23** or **4.3 times faster** than cpython

I prefer to write "4.3x faster" than "0.23 (faster)".

This PR is also a reply to this tweet of James Abel:

> I can review. One thing I prefer is to have > 1.0 always mean faster. For example say 0.8x instead of 1.25x slower. Also I like to avoid percentages if possible (it seems like they make it into marketing materials anyway :) )

https://twitter.com/JamesAbel/status/1321320012586442752

jamesabel:
A couple of things here.  These are all my opinion (although influenced by the performance engineers I worked with at Intel :) )

Performance charts should provide the overall measurement units, e.g. are they in time, frames per second, number of simultaneous users, etc. so that the reader knows if lower or higher is better for a *measurement*.  In general for a given set of measurements the units are the same (you can have scaling differences like mS and uS, but don't mix seconds and FPS).  This is what you are doing, and this allows for comparisons.

Then when *comparing* measurements (e.g. the geomean) > 1.0 is always better for the unit (or condition) under test vs. the baseline.  (For the sake be being explicit, often somewhere in the comparison table "higher is better" is stated.)

This way you won't have a mix in a table of "slower" and "faster".  Everything is a ratio and higher is better (in your case faster).  This way you don't have to state "slower" or "faster".  Less than 1.0 is always worse (slower) and greater than 1.0 is always better (faster).

Therefore, for your numbers above you'd have 1.09x and 0.82x (as opposed to 1.22x), given that the former is better and the latter is worse.

It also makes the code easier - no conditionals. :)

Sometimes "marketing" will explicitly add a "faster" or "slower" label and/or coloring (red=slower, green=faster, yellow=same), but that's cosmetic (albeit sometimes effective).

I hope this helps.

vstinner:
@jamesabel: Thanks for your feedback! pyperf stores the unit of a benchmark (seconds, most of the time) and displays it in show and compare_to commands.

About "0.23 (faster)" vs "4.3x faster", it may depend on your audience. In my experience, most people misunderstand benchmark results. I tried in pyperf to make the output to reduce the risk of understanding a result backwards (consider that a change makes Python faster, rather it makes Python slower).

I understand the argument about consistency, but let's say that pyperf is opinionated and I prefer to only display numbers >= 1.0 for ratio (faster or slower and geometric means).

I merged manually my perf as https://github.com/psf/pyperf/commit/97618bb4ba5bb57f43e9859947cf3780c6b500b6

I also removed percentage to reduce number displayed to readers, again, to reduce the risk of confusion.

Issue #92, opened 2021-03-12
Support --track-memory option on macOS


corona10:
`--track-memory` does not support macOS yet.
First I'd like to start this issue by using `psutil` any ideas?

corona10:
cc @vstinner

PR #93, opened 2021-03-12
Support --track-memory option on macOS

corona10:
closes: #92


pyperf/_psutil_memory.py
```diff
@@ -0,0 +1,44 @@
+try:
+    import os
+    import psutil
+    import threading
+    import time
+except ImportError:
+    raise ImportError('psutil is not installed')
```
vstinner, line 7: Why not pass through the error on "import psutil"?

You may update/clarify when psutil is needed at:
https://pyperf.readthedocs.io/en/latest/run_benchmark.html#install-pyperf

pyperf/_psutil_memory.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class PeakMemoryUsageThread(threading.Thread):
+    def __init__(self):
+        threading.Thread.__init__(self)
+        self.process = psutil.Process(os.getpid())
+        self.peak_usage = 0
+        self._done = threading.Event()
+        self.sleep = 0.001   # 1 ms
```
vstinner, line 16: It may use too much CPU time, a worker process runs at least 300 ms. Sampling every 10 ms should be enough.

Oh. pyperf/_memory.py also uses 1 ms. Maybe it should be increased as well :-)

What do you think?

pyperf/_psutil_memory.py
```diff
@@ -0,0 +1,44 @@
...
+    import threading
+    import time
+except ImportError:
+    raise ImportError('psutil is not installed')
+
+
+class PeakMemoryUsageThread(threading.Thread):
+    def __init__(self):
+        threading.Thread.__init__(self)
+        self.process = psutil.Process(os.getpid())
```
vstinner, line 13: I'm not used to the psutils.Process API. Is there a close() method when you are done with it?

pyperf/_runner.py
```diff
@@ -310,6 +310,8 @@ def _process_args_impl(self):
         if args.track_memory:
             if MS_WINDOWS:
                 from pyperf._win_memory import check_tracking_memory
+            elif MAC_OS:
+                from pyperf._psutil_memory import check_tracking_memory
             else:
                 from pyperf._memory import check_tracking_memory
```
vstinner, line 316: Maybe pyperf/_memory.py should be a portable API which switchs between the Linux, Windows and macOS implementation. (And _memory.py should be renamed to _linux_memory.py)

pyperf/_worker.py
```diff
@@ -310,6 +310,10 @@ def compute(self):
         if args.track_memory:
             if MS_WINDOWS:
                 from pyperf._win_memory import get_peak_pagefile_usage
+            elif MAC_OS:
+                from pyperf._psutil_memory import PeakMemoryUsageThread
+                mem_thread = PeakMemoryUsageThread()
+                mem_thread.start()
             else:
                 from pyperf._memory import PeakMemoryUsageThread
                 mem_thread = PeakMemoryUsageThread()
```
vstinner, line 319: It would be interesting to have an abstraction to hide the OS-specific implementation, but if you don't want to do it, you leave it as it is ;-)

pyperf/tests/test_timeit.py
```diff
@@ -220,6 +220,9 @@ def test_cli_snippet_error(self):
     # When the PyPy program is copied, it fails with "Library path not found"
     @unittest.skipIf(pyperf.python_implementation() == 'pypy',
                      'pypy program cannot be copied')
+    # FIXME: macOS fail this test
```
vstinner, line 223: any idea what's going on?

pyperf/_worker.py
corona10: Let's do it as a seperate issue ;)

pyperf/tests/test_timeit.py
corona10: Not digging yet, but soon I will find out what's going on

pyperf/_psutil_memory.py
corona10: AFAIK, no close method for this API

pyperf/_psutil_memory.py
corona10: Agree, I've updated the sampling period as 10ms ;)

pyperf/_psutil_memory.py
corona10: Done!

.github/workflows/build.yml
```diff
@@ -8,11 +8,12 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        os: [ubuntu-latest, macos-latest]
```
corona10, line 13: @vstinner 
We may able to add `windows-latest` but I didn't check pyperf test on the Windows machine.
Did you ever been to run tests on Windows?

vstinner: If there is a reliable Windows CI, why not. But for my own usage, Linux support is enough. I only try to provide best support for macOS and Windows.

pyperf/_linux_memory.py
```diff
@@ -29,7 +29,7 @@ def __init__(self):
         threading.Thread.__init__(self)
         self.peak_usage = 0
         self._done = threading.Event()
-        self.sleep = 0.001   # 1 ms
+        self.sleep = 0.01   # 10 ms
```
vstinner, line 32: I like to write 3 digits to align to milliseconds :-p it's just a personal preference.

```suggestion
        self.sleep = 0.010   # 10 ms
```

pyperf/_psutil_memory.py
vstinner: Ok, I was just curious.

pyperf/tests/test_timeit.py
vstinner: I would prefer to not merge this change. Otherwise, the test may be skipped forever and nobody will notice :-(

corona10:
@vstinner 
- Reflect code review
- Remove macOS CI until all tests are passed ;)

vstinner (approved):
LGTM, but I didn't test the macOS code, I trust you ;-)

You may run some manual tests on Linux. I don't think that there are automated tests yet.

PR #98, opened 2021-03-16
Fix memory unit for macOS when calculating ru_maxrss

corona10:
## AS-IS
```
(.tmp) ➜  pyperf git:(main) ✗ python -m pyperf command --track-memory python -c pass
.....................
command: Mean +- std dev: 5288.8 MB +- 60.9 MB
```
## TO-BE
```
(.tmp) ➜  pyperf git:(main) ✗ python -m pyperf command --track-memory python -c pass
.....................
command: Mean +- std dev: 5315.4 kB +- 75.8 kB
```

### DIFF
--- a/pyperf/_process_time.py
+++ b/pyperf/_process_time.py
@@ -28,6 +28,8 @@ except ImportError:
 def get_max_rss():
     if resource is not None:
         usage = resource.getrusage(resource.RUSAGE_CHILDREN)
+        if sys.platform == 'darwin':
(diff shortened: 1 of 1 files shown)


vstinner (approved):
LGTM.

I'm not sure why you cannot merge a PR if I don't approve it.

vstinner:
> command: Mean +- std dev: 5288.8 MB +- 60.9 MB

Wow, that's a lot :-D

Issue #121, opened 2022-01-12
timeit support for coroutines


Tinche:
Hello!

I think `pyperf` is an amazing project and I use the `timeit` to benchmark essentially all the libraries I work on (attrs, cattrs, incant...).

I wish I could use it to benchmark async functions though. Right now, I benchmark `asyncio.run(my_coro)` but since `asyncio.run` is so costly there's a ton of noise in the signal.

I think essentially pyperf could detect a coroutine was passed in, spawn an event loop and just await it in a loop.

corona10:
@vstinner Do you have any ideas?

vstinner:
I don't know how to do that.

cc @methane

methane:
@Tinche Do you really mean `timeit`, but not `bench_func()`.

`bench_func()` receives function. So it may be possible to detect the func is coroutine or not.
On the other hand, `timeit` receives an expression, not a function. So it is difficult to detect that coroutine is passed in.

methane:
Would you give us some examples?

Tinche:
I use `timeit` all the time in the terminal and I've never used `bench_func()`, so probably `timeit`. Maybe it could be a flag or a different command?

Here's an example. I have a project, https://github.com/Tinche/incant/, that does function composition (mostly for dependency injection), and I want to measure how efficient it is. It supports functions and coroutines. Functions I can benchmark easily, coroutines I need to benchmark using `asyncio.run`, and that has a ton of noise since it does a lot of unrelated work.

Note that these coroutines I'm benchmarking are technically async, but they either do not await anything or they await `sleep(0)`.

I usually prepare the function being tested in a file and then do something like:
```
pyperf timeit -g -s "from asyncio import run; from test import main" "run(main())"
```
so since I need to have a separate file `bench_func()` could work too. The CLI interface is sooo nice though ;)

That said, maybe there's a way to run a coroutine without involving an event loop? Just iterate over it until it's done or something like that? I'm not proficient in that part of Python.

methane:
Would you try this?

```py
pyperf timeit -g -s "import asyncio; loop=asyncio.get_event_loop(); from test import main" "loop.run_until_complete(main())"
```

With this, one loop is used repeatedly instead of creating and destroying loops for each main() execution.
Is this reduce your "noise"?

Tinche:
It does work and helps a little. If it's too hard to do otherwise in pyperf I will accept this as the answer ;)

methane:
[...]

PR #124, opened 2022-02-03
Add bench_async_func()

methane:
Ref #121.


pyperf/_runner.py
```diff
@@ -496,6 +496,43 @@ def task_func(task, loops):
...
+            else:
+                t0 = local_timer()
+                loop.run_until_complete(local_func())
+                dt = local_timer() - t0
+
+            return dt
+
+        task = WorkerProcessTask(self, name, task_func, metadata)
+        task.inner_loops = inner_loops
+        return self._main(task)
```
vstinner, line 534: How is this function better than ``runner.bench_func(loop.run_until_complete, func)``? Do you propose to add it because it is more convenient, or because you expect a better accuracy?

vstinner: From https://github.com/psf/pyperf/issues/121#issuecomment-1028484827 comment, I understand that the problem of `bench_func()` is the temptation to use it with `asyncio.run()` which adds the large overhead of creating a new event loop at each iteation.

doc/api.rst
```diff
@@ -538,6 +539,21 @@ Runner class
 
       See the :ref:`bench_func() example <bench_func_example>`.
 
+   .. method:: bench_async_func(name, func, \*args, inner_loops=None, metadata=None)
+
+      Benchmark the function ``await func(*args)`` in asyncio event loop.
```
vstinner, line 544: Compared to ``asyncio.run()``, the risk is that func() coroutine leaves pending events which will be handled by the following benchmark, or between iterations of the same benchmark value. I'm thinking at sockets and loop.call_later().

Should the documentation warns about these issues? The developer should pay attention to no "leak" any pending event in func() coroutine.

pyperf/_runner.py
```diff
@@ -496,6 +496,43 @@ def task_func(task, loops):
...
+
+        if args:
+            func = functools.partial(func, *args)
+
+        def task_func(task, loops):
+            import asyncio
+            # use fast local variables
+            local_timer = time.perf_counter
+            local_func = func
+            loop = asyncio.get_event_loop()
```
vstinner, line 517: Would it make sense to attempt reimplementing asyncio.run() here, but only measure run_until_complete() timing? I'm talking about creating a fresh event loop each time, and shutdown asynchronous generators and the default executor. Well, same things than asyncio.run().

methane: I use asyncio.run() instead of reimplement it.

doc/examples/bench_async_func.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
```
vstinner, line 1: Can you mark the file as executable? (file permissions)

vstinner (commented):
Can you please add a changelog entry in doc/changelog.rst? You need to add Version 2.3.1 section.

vstinner:
I tested the example on my Fedora 35 with Python 3.10. I get:

```
async_sleep: Mean +- std dev: 1.32 ms +- 0.02 ms
```

That's interesting :-) Using ``await asyncio.sleep(1e-6)``, I get:

```
async_sleep: Mean +- std dev: 13.0 us +- 0.6 us
```

vstinner:
Oh, sadly ``asyncio.run()`` doesn't exist in Python 3.6. Is it possible to reimplement ``asyncio.run()`` in pyperf on Python 3.6?

pyperf currently still supports Python 3.6: https://pyperf.readthedocs.io/en/latest/run_benchmark.html#install-pyperf

pyperf/_runner.py
```diff
@@ -496,6 +496,59 @@ def task_func(task, loops):
...
+                    local_timer = time.perf_counter
+                    local_func = func
+
+                    t0 = local_timer()
+                    await local_func()
+                    dt = local_timer() - t0
+                    return dt
+
+            import asyncio
+            try:
```
vstinner, line 537: I would prefer to testing ``hasattr(asyncio, 'main')``, since main() can raise AttributeError.

pyperf/_runner.py
```diff
@@ -496,6 +496,59 @@ def task_func(task, loops):
...
+                dt = asyncio.run(main())
+            except AttributeError:  # Python 3.6 doesn't have asyncio.run
+                loop = asyncio.new_event_loop()
```
vstinner, line 540: I suggest to add ``asyncio.set_event_loop(loop)`` here.

pyperf/_runner.py
```diff
@@ -496,6 +496,59 @@ def task_func(task, loops):
...
+                try:
+                    dt = loop.run_until_complete(main())
+                finally:
+                    loop.close()
```
vstinner, line 544: And add:

```
asyncio.set_event_loop(None)
```

Note: asyncio only started to wait for the default executor in ``asyncio.run()`` in Python 3.9.

doc/api.rst
```diff
@@ -538,6 +539,25 @@ Runner class
...
+      The *inner_loops* parameter is used to normalize timing per loop
+      iteration.
+
+      To call ``func()`` with keyword arguments, use ``functools.partial``.
+
+      Return a :class:`Benchmark` instance.
+
+      .. note::
+         If the ``func()`` leaves any tasks, it affects to next iteration.
+         So you should care about task leak for reliable benchmark.
```
vstinner, line 557: This note is now outdated, since a new event loop is created at each iteration, no?

vstinner:
With the proposed implementation (measure time inside asyncio.run and not outside), I agree that the implementation increases the accuracy and so it's worth it to add a new method ;-)

vstinner (approved):
LGTM, thanks for the updates!

vstinner:
I completed the changelog in commit 0814a4af62646b00cb2908db67f873c5910dd913.

vstinner:
This change is now part of the just released pyperf 2.3.1.

PR #132, opened 2022-05-24
Support reporting geometric mean by benchmark tags

mdboom:
Addresses https://github.com/python/pyperformance/issues/208

This reports geometric mean organized by the tag(s) assigned to each benchmark.
This will allow us to include benchmarks in the pyperformance suite that we
don't necessarily want to include in "one big overall number" to represent progress.

### DIFF
--- a/pyperf/_metadata.py
+++ b/pyperf/_metadata.py
@@ -42,6 +42,12 @@ def is_positive(value):
     return (value >= 0)
 
 
+def is_tags(value):
+    if not isinstance(value, list):
+        return False
+    return all(isinstance(x, str) and x not in ('all', '') for x in value)
+
+
 def parse_load_avg(value):
     if isinstance(value, NUMBER_TYPES):
         return value
@@ -62,6 +68,7 @@ DATETIME = _MetadataInfo(format_noop, (str,), None, None)
 LOOPS = _MetadataInfo(format_number, (int,), is_strictly_positive, 'integer')
 WARMUPS = _MetadataInfo(format_number, (int,), is_positive, 'integer')
 SECONDS = _MetadataInfo(format_seconds, NUMBER_TYPES, is_positive, 'second')
+TAGS = _MetadataInfo(format_generic, (list,), is_tags, 'tag')
 
 # Registry of metadata keys
 METADATA = {
@@ -84,6 +91,7 @@ METADATA = {
     'recalibrate_loops': LOOPS,
     'calibrate_warmups': WARMUPS,
     'recalibrate_warmups': WARMUPS,
+    'tags': TAGS,
 }
 
 DEFAULT_METADATA_INFO = _MetadataInfo(format_generic, METADATA_VALUE_TYPES, None, None)
--- a/pyperf/tests/test_perf_cli.py
+++ b/pyperf/tests/test_perf_cli.py
@@ -330,6 +330,83 @@ class TestPerfCLI(BaseTestCase, unittest.TestCase):
         """
         self.check_command(expected, 'compare_to', '--table', "--group-by-speed", py36, py37)
 
+    def test_compare_to_cli_tags(self):
+        py36 = os.path.join(TESTDIR, 'mult_list_py36_tags.json')
+        py37 = os.path.join(TESTDIR, 'mult_list_py37_tags.json')
+
+        # 2 files
+        expected = """
+            Benchmarks with tag 'bar':
+            ==========================
+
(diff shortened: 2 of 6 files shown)


pyperf/_metadata.py
```diff
@@ -62,6 +68,7 @@ def format_noop(value):
 LOOPS = _MetadataInfo(format_number, (int,), is_strictly_positive, 'integer')
 WARMUPS = _MetadataInfo(format_number, (int,), is_positive, 'integer')
 SECONDS = _MetadataInfo(format_seconds, NUMBER_TYPES, is_positive, 'second')
+TAGS = _MetadataInfo(format_generic, (tuple, list), is_tags, 'tag')
```
vstinner, line 71: It's annoying to have to deal with two different types, tuple or list. I would prefer to always use the same type. A tuple should be good for tags, no?

pyperf/_metadata.py
```diff
@@ -42,6 +42,12 @@ def is_positive(value):
     return (value >= 0)
 
 
+def is_tags(value):
+    if not isinstance(value, (list, tuple)):
+        return False
+    return all(isinstance(x, str) and x != 'all' for x in value)
```
vstinner, line 48: I suggest to reject empty tags. They would be annoying for users, like specifying an empty tag on the command line.

vstinner:
It seems like you want to add a new key in metadata. In that case, you should specify it in the doc: https://pyperf.readthedocs.io/en/latest/api.html#metadata

I dislike changing the default formatting to add "(all)". Can you try to omit "(all)"? For example, if no benchmark has tags, it's weird to display the magic "all" tag.

```
            +----------------------+---------------------+-----------------------+
            | Benchmark            | mult_list_py36_tags | mult_list_py37_tags   |
            +======================+=====================+=======================+
            | [1]*1000             | 2.13 us             | 2.09 us: 1.02x faster |
            +----------------------+---------------------+-----------------------+
            | [1,2]*1000           | 3.70 us             | 5.28 us: 1.42x slower |
            +----------------------+---------------------+-----------------------+
            | [1,2,3]*1000         | 4.61 us             | 6.05 us: 1.31x slower |
            +----------------------+---------------------+-----------------------+
            | Geometric mean (all) | (ref)               | 1.22x slower          |
            +----------------------+---------------------+-----------------------+
            | Geometric mean (bar) | (ref)               | 1.37x slower          |
            +----------------------+---------------------+-----------------------+
            | Geometric mean (foo) | (ref)               | 1.18x slower          |
            +----------------------+---------------------+-----------------------+
```

In this table, it's also not easy for me to understand which benchmarks are used to compute the geometric means for each tag, since benchmark tags are not listed. Would it make sense to list tags?

vstinner:
Do you have real examples of tags on benchmarks? I mean what are real tag values?

vstinner:
> In this table, it's also not easy for me to understand which benchmarks are used to compute the geometric means for each tag, since benchmark tags are not listed. Would it make sense to list tags?

Another option is to render one table per tag: it would only list benchmarks matching this tag, and so the "geometric mean" final row would summarize the table. And there would always be a last table with all benchmarks.

mdboom:
@vstinner: Do these changes work for you?

pyperf/tests/test_perf_cli.py
```diff
@@ -330,6 +330,83 @@ def test_compare_to_cli(self):
...
+            +================+=====================+=======================+
+            | [1]*1000       | 2.13 us             | 2.09 us: 1.02x faster |
+            +----------------+---------------------+-----------------------+
+            | [1,2]*1000     | 3.70 us             | 5.28 us: 1.42x slower |
+            +----------------+---------------------+-----------------------+
+            | [1,2,3]*1000   | 4.61 us             | 6.05 us: 1.31x slower |
+            +----------------+---------------------+-----------------------+
+            | Geometric mean | (ref)               | 1.22x slower          |
+            +----------------+---------------------+-----------------------+
+        """
```
vstinner, line 407: Oh wow, that looks great, thank you!

vstinner (approved):
LGTM.

vstinner:
Sadly, the tests fail:

```
- FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/pyperf/pyperf/pyperf/tests/mult_list_py36_tags.json'
```

mdboom:
> Sadly, the tests fail:
> 
> ```
> - FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/pyperf/pyperf/pyperf/tests/mult_list_py36_tags.json'
> ```

Sorry -- forgot to commit the new test files -- let's see how this works.

Issue #133, opened 2022-06-03
Support collecting cProfile data from a benchmark


mdboom:
While this wouldn't be useful for absolute timing (benchmarking), it's sometimes handy to generate a profile from a benchmark to understand where the time is being spent.  It's currently possible to do this by hacking up a benchmark script, but it would be better to make it as simple as:

```
python run_benchmark.py --profile output.profile
```

sergey-miryanov:
Since we now have a hook for Tachyon (https://github.com/psf/pyperf/pull/230), can this be closed?

vstinner:
Ok, I close this very old ticket!

PR #134, opened 2022-06-03
Add support for profiling benchmarks

mdboom:
This should make it much easier to collect profiles for benchmarks in the pyperformance suite.

Implements #133.

### DIFF
--- a/doc/cli.rst
+++ b/doc/cli.rst
@@ -491,6 +491,7 @@ Usage
         [--duplicate DUPLICATE]
         [-s SETUP]
         [--teardown TEARDOWN]
+        [--profile PROFILE]
         stmt [stmt ...]
 
 Options:
@@ -512,6 +513,7 @@ Options:
 * ``--duplicate=DUPLICATE``: Duplicate statements (``stmt`` statements, not
   ``SETUP``) to reduce the overhead of the outer loop and multiply
   inner loops by DUPLICATE (see ``--inner-loops`` option).
+* ``--profile=PROFILE``: Run the benchmark inside the cProfile profiler and output to the given file. This is a convenient way to profile a specific benchmark, but it will make the actual benchmark timings much less accurate.
 
 .. note::
    timeit ``-n`` (number) and ``-r`` (repeat) options become ``-l`` (loops) and
--- a/pyperf/_manager.py
+++ b/pyperf/_manager.py
@@ -66,6 +66,9 @@ class Manager(object):
         if args.track_memory:
             cmd.append('--track-memory')
 
+        if args.profile:
+            cmd.extend(['--profile', args.profile])
+
         if self.runner._add_cmdline_args:
             self.runner._add_cmdline_args(cmd, args)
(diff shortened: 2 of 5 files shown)


vstinner:
Tests don't pass.

How does merge_profile_stats() work? Does it compute the average of N processes timings? Or does it accumulate time?

Is it more reliable than running a single process?

mdboom:
> Tests don't pass.

Noted. Wanted to get some feedback about the general concept here first.

> How does merge_profile_stats() work? Does it compute the average of N processes timings? Or does it accumulate time?
> 
> Is it more reliable than running a single process?

Yes, it just accumulates timings from multiple profiling runs. It produces more accurate results, in the same way that running the benchmarks multiple times does.

mdboom:
I'm not sure I understand the request.

The temporary file only comes into play when using `bench_process`.  In that case, the temporary filename is generated in parent process and passed to the child process being benchmarked specifically to make it easier to clean up because it could crash.  (The temporary file is deleted whether the worker process succeeds or fails).

For other kinds of benchmarks, there is no temporary file involved -- each child worker process merges their results directly into the output file.  That has a separate problem in that there is a race condition if multiple workers update that file at the same time, but the whole design here is to not benchmark things in parallel, so that should be ok.

This certainly could use a pipe to communicate all the profiling data to the parent process -- all platforms already use that to communicate benchmark results from the worker processes.  But it would add complexity to a bunch of places since the "protocol", which right now all dumps things directly into the master benchmarking results, would have to split things out into separate files for the profiling results.

vstinner:
cc @corona10

vstinner:
cc @pablogsal

corona10:
I will left review by this weekend cc @vstinner

pyperf/tests/test_runner.py
```diff
@@ -73,7 +74,7 @@ class ClockInfo:
         self.assertEqual(bench.get_nrun(), 1)
 
         return Result(runner, bench, stdout)
-
+
```
corona10, line 77: Please remove blank

```suggestion

```

corona10 (changes requested):
Please update the following documentation.

https://github.com/psf/pyperf/blob/6eb6de7427c5ac8939a21c0c68014be1167847de/doc/cli.rst#pyperf-timeit

corona10:
I am going to release the new version of pyperf in 7days after this PR is merged.
cc @vstinner

mdboom:
> Please update the following documentation.
> 
> https://github.com/psf/pyperf/blob/6eb6de7427c5ac8939a21c0c68014be1167847de/doc/cli.rst#pyperf-timeit

@corona10: I already did that in this PR. Is there something specific missing there that you'd like to see?

corona10 (approved):
@mdboom 

> @corona10: I already did that in this PR. Is there something specific missing there that you'd like to see?

Oops sorry, I may miss something at that time.
LGTM, I will release the next version by this weekend.

PR #139, opened 2022-08-25
Recognize new name of Graal-based Python as jitted

timfel:
`sys.implementation.name` was changed in recent Graal nightlies.

### DIFF
--- a/pyperf/_utils.py
+++ b/pyperf/_utils.py
@@ -205,7 +205,7 @@ def python_has_jit():
     implementation_name = python_implementation()
     if implementation_name == 'pypy':
         return sys.pypy_translation_info["translation.jit"]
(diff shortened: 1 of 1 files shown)


ericsnowcurrently (approved):

corona10 (approved):
lgtm

timfel:
thank you!

vstinner:
Defining if a Python implementation has a JIT or not became complicated. CPython 3.10 has an "opcode cache". CPython 3.11 now has "adaptative bytecode" which changes the bytecode after a few iterations.

It might help to implement a *changepoint detection* to decide if a benchmark looks steady or not.

My notes: https://vstinner.readthedocs.io/pypy_warmups.html

PR #140, opened 2022-10-13
Collect specialization statistics from running benchmarks

mdboom:
This will collect statistics on CPython builds with `--enable-pyperf`, and only during the run of the benchmarks themselves.

Cc @markshannon

### DIFF
--- a/doc/run_benchmark.rst
+++ b/doc/run_benchmark.rst
@@ -191,4 +191,14 @@ mentioned above. Running an arbitrary number of warmup values may also make
 the benchmark less reliable since two runs may use a different number of warmup
 values.
 
+Specializer statistics (`pystats`)
+==================================
 
+If running benchmarks on a CPython built with the ``--enable-pystats`` flag, pyperf will automatically collect statistics from about the bytecode specializer during the run of benchmark code.
+
+Due to the overhead of collecting the statistics, it is unlikely the timing results will be useful.
+
+The `Tools/scripts/summarize_stats.py <https://github.com/python/cpython/blob/main/Tools/scripts/summarize_stats.py>`_ script to summarize the statistics in a human-readable form.
+
+Statistics are not cleared between runs.
+If you need to delete statistics from a previous run, remove the files in `/tmp/py_stats` (Unix) or `C:\temp\py_stats` (Windows).
--- a/pyperf/__init__.py
+++ b/pyperf/__init__.py
@@ -18,3 +18,12 @@ __all__.extend(('Run', 'Benchmark', 'BenchmarkSuite', 'add_runs'))
 
 from pyperf._runner import Runner   # noqa
 __all__.append('Runner')
+
+import sys
+
+# Reset the stats collection if running a --enable-pystats build
+try:
+    sys._stats_off()
+    sys._stats_clear()
+except AttributeError:
+    pass
--- a/pyperf/_collect_metadata.py
+++ b/pyperf/_collect_metadata.py
@@ -116,6 +116,10 @@ def collect_python_metadata(metadata):
         if not gc.isenabled():
             metadata['python_gc'] = 'disabled'
 
+    # pystats enabled?
+    if hasattr(sys, "_stats_clear"):
+        metadata['python_pystats'] = 'enabled'
+
 
 def read_proc(path):
     path = proc_path(path)
--- a/pyperf/_runner.py
+++ b/pyperf/_runner.py
@@ -62,6 +62,24 @@ def profiling_wrapper(func):
     return profiler, profiling_func
 
 
+def stats_wrapper(func):
+    """
+    Wrap a function to turn on/off stats collection.
+
+    If this isn't an `--enable-pystats` build, the original function is returned
+    and has no overhead.
+    """
+    if hasattr(sys, "_stats_on"):
+        def stats_func(*args):
+            sys._stats_on()
+            result = func(*args)
+            sys._stats_off()
+            return result
+        return stats_func
+    else:
+        return func
+
+
 class CLIError(Exception):
     pass
 
@@ -475,6 +493,8 @@ class Runner:
         if not self._check_worker_task():
             return None
 
+        time_func = stats_wrapper(time_func)
+
         if self.args.profile:
             profiler, time_func = profiling_wrapper(time_func)
 
@@ -500,6 +520,8 @@ class Runner:
 
         if not self._check_worker_task():
             return None
+
+        func = stats_wrapper(func)
 
         if args:
             func = functools.partial(func, *args)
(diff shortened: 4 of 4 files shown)


pyperf/__init__.py
```diff
@@ -18,3 +18,12 @@
 
 from pyperf._runner import Runner   # noqa
 __all__.append('Runner')
+
+import sys
+
+# Reset the stats collection if running a --enable-pystats build
+try:
+    sys._stats_off()
+    sys._stats_clear()
```
vstinner, line 27: Please don't execute code on "import pyperf". It should be a deliberate action to clear such cache.

pyperf/_runner.py
```diff
@@ -501,6 +522,8 @@ def bench_func(self, name, func, *args, **kwargs):
         if not self._check_worker_task():
             return None
 
+        func = stats_wrapper(func)
```
vstinner, line 525: On a Python built with sys._stats_on(), this wrapper adds a fixed overhead. If you want to execute code before/after the benchmark, it should be done before/after code measuring time.

See task_func() below.

vstinner (changes requested):
> If running benchmarks on a CPython built with the ``--enable-pystats`` flag, pyperf will automatically collect statistics from about the bytecode specializer during the run of benchmark code.

I don't see any code to store statistics, so I don't get the purpose of this change. pyperf spawns many subprocesses, how are statistics collected? Do you compute the average? I don't get it.

mdboom:
The statistics are collected in the specializing optimizer in the interpreter loop in CPython itself when built with `--enable-pystats`.  The statistics are dumped in separate files at process shutdown in `/tmp/py_stats` (which is how this works with multiple processes).  You can see some example output of the `summarize_stats.py` script here: https://github.com/faster-cpython/ideas/blob/main/stats.md

pyperf/__init__.py
mdboom: I can probably get away with putting this in the `Runner` constructor -- it's possible, I suppose, to have more than one of those, but in practice that doesn't happen in the pyperformance benchmarks.

And to be clear, this doesn't affect anything on disk -- it is only the in-memory statistics collected so far.  (This is basically to remove the effect of the statistics collected during pyperf's startup and book-keeping).

pyperf/_runner.py
mdboom: In this case, the benchmarking timings are useless anyway (due to the stats overhead), but it would have a (minor) impact on the statistics themselves.  I'll see if I can refactor this, though.  It probably requires a copy-and-paste of `task_func` to really minimize the overhead -- I'm not sure it's worth it, to be honest.

markshannon: If we are gathering stats, the timings are meaningless.
To maximize the accuracy of the stats, we want to turn them on and off as close to the function call as possible.
This looks about as good as we can get.

markshannon (approved):

doc/run_benchmark.rst
```diff
@@ -191,4 +191,14 @@ mentioned above. Running an arbitrary number of warmup values may also make
 the benchmark less reliable since two runs may use a different number of warmup
 values.
 
+Specializer statistics (`pystats`)
+==================================
 
+If running benchmarks on a CPython built with the ``--enable-pystats`` flag, pyperf will automatically collect statistics from about the bytecode specializer during the run of benchmark code.
```
vstinner, line 197: I dislike "will automatically collect statistics". pyperf usually puts everything in the JSON file, but here all it does is to call ``sys._stats_on()`` before and ``sys._stats_off()`` after. It doesn't clear existing statistics in /tmp/py_stats, it doesn't store the in the JSON file, it doesn't compute average or anything.

Would you mind to mention ``sys._stats_on()`` and ``sys._stats_off()`` functions instead?

vstinner (commented):
@markshannon:
> If we are gathering stats, the timings are meaningless. To maximize the accuracy of the stats, we want to turn them on and off as close to the function call as possible. This looks about as good as we can get.

Hum, ok. I'm not against this change in this case, but I still have some concerns about the documentation of this change.

doc/run_benchmark.rst
```diff
@@ -191,4 +191,14 @@ mentioned above. Running an arbitrary number of warmup values may also make
...
+
+Due to the overhead of collecting the statistics, it is unlikely the timing results will be useful.
```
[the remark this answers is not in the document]
mdboom: Sure, that's fine.

mdboom:
Anything else you'd like to see on this, @vstinner?

mdboom:
I'm converting this to a draft -- on full A/A testing, it turns out the results from this a quite non-deterministic.  I think the noise is coming from the warmup / calibration stages, which don't always run the same number of times.  I *think* by excluding these runs, we'll get more stable results, but checking that over [here](https://github.com/mdboom/cpython/pull/3) first.

mdboom:
This is ready for review.  With this change, we get consistent results running the pyperformance benchmark suite multiple times.  See this [prototype](https://github.com/mdboom/cpython/pull/3).

doc/run_benchmark.rst
```diff
@@ -191,4 +191,15 @@ mentioned above. Running an arbitrary number of warmup values may also make
 the benchmark less reliable since two runs may use a different number of warmup
 values.
 
+Specializer statistics (``pystats``)
+==================================
 
+``pyperf`` has built-in support for `specializer statistics (``pystats``) <https://docs.python.org/dev/using/configure.html#cmdoption-enable-pystats>`_.
+If running benchmarks on a CPython built with the ``--enable-pystats`` flag, pyperf will automatically collect ``pystats`` on the benchmark code by calling ``sys._stats_on`` immediately before the benchmark and calling ``sys._stats_off`` immediately after.
```
vstinner, line 198: Since you go into details, you should also mention that they start by calling ``_stats_clear()``. Moreover, you mention "only if we aren't warming up or calibrating".

mdboom: I added "Stats are not collected when running ``pyperf``'s own code or when warming up or calibrating the benchmarks."  I don't think mentioning `_stats_clear()` is necessary -- the point of that is to not include `pyperf`'s own code.

pyperf/_collect_metadata.py
```diff
@@ -116,6 +116,10 @@ def collect_python_metadata(metadata):
         if not gc.isenabled():
             metadata['python_gc'] = 'disabled'
 
+    # pystats enabled?
+    if hasattr(sys, "_stats_clear"):
+        metadata['python_pystats'] = 'enabled'
```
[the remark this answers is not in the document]
mdboom: Makes sense.

vstinner (approved):
LGTM. Thanks for the multiple updates ;-)

PR #142, opened 2022-11-04
Add support for Python 3.11

hugovk:
Also drop support for EOL Python <= 3.6 and fix errors found running `tox -e pep8`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,6 @@ jobs:
         python: ['3.11']
         include:
           - os: ubuntu-latest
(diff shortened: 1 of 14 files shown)


corona10 (approved):
lgtm

hugovk:
It removes a CI job, but that's not a big deal here. 

I didn't upgrade the syntax for 3.7+, so it's fine to re-add it.

vstinner:
It's not a big deal. I'm also fine with removing Python 3.6 support. Especially, if technically Python 3.6 still works :-D

Issue #146, opened 2022-11-21
Propagating PYTHONPATH in tests


tjni:
I am working on running the test suite of this package in [nixpkgs](https://github.com/NixOS/nixpkgs).

The nix package manager builds each Python package from source and installs it into its own location in a read-only directory instead of relying on a virtual environment and a wheel. To build an environment with dependencies in which a package can be built, we figure out where the dependencies are in the read-only directory and add them to PYTHONPATH.

Currently, a number of tests are failing because they spawn subprocesses which do not have PYTHONPATH passed along, so dependencies of pyperf such as psutil will not be found.

Would you accept a PR to inherit this environment variable in tests, or do you have other suggestions?

vstinner:
@corona10 @pablogsal: I'm fine with inheriting PYTHONPATH by default, or even consider inheriting all environment variables by default. What do you think?

corona10:
> @corona10 @pablogsal: I'm fine with inheriting PYTHONPATH by default, or even consider inheriting all environment variables by default. What do you think?

I can not expect the side effect, so looks good to go.

PR #148, opened 2022-11-21
Inherit PYTHONPATH when spawning workers.

tjni:
Resolves https://github.com/psf/pyperf/issues/146.


vstinner (commented):
Would you mind to document the change in doc/changelog.rst?

You have to increase the pyperf version and add a new section. setup.py says:

> update version in setup.py, pyperf/__init__.py and doc/conf.py

vstinner:
Ah, mention also that PYTHONPATH is inherited in ``--inherit-environ`` doc: doc/runner.rst.

vstinner:
Merged, thanks.

Issue #115, opened 2021-09-28
Migrate to the pyproject.toml


corona10:
The title says it all

PR #149, opened 2022-11-27
gh-115: Migrate to the pyproject.toml

corona10:
closes: #115


pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+#  - monitor the publish status: https://github.com/psf/pyperf/actions/workflows/publish.yml
+#
+# After the release:
+#
+#  - set version to n+1
+#  - git commit -a -m "post-release"
+#  - git push or send the PR to the repository
+
+[build-system]
+requires = ["setuptools >= 35.0.2", "setuptools_scm >= 2.0.0, <3", "wheel"]
```
CAM-Gerlach, line 26:
```suggestion
requires = ["setuptools >= 61"]
```

Fixed several issues here:

* Setuptools >=61 is required, at minimum, to use Pyproject metadata as you do here
* `wheel` is not a direct build dependency, so it should just be removed
* Setuptools_scm is not, as far as I can tell, used anywhere by this project, either in the old `setup.py` (which contains no `use_scm_version=True` or the new `pyproject.toml` (which contains no `[tool.setuptools_scm]` section), and in fact uses a checked-in version. Furthermore 2.0.0 is nearly 5 years old, and 3.0 over 4 years old, so it wouldn't work properly with `pyproject.toml` or other modern standards anyway. Therefore, I removed it

.github/workflows/publish.yml
```diff
@@ -16,10 +16,10 @@ jobs:
         python-version: '3.x'
     - name: Install dependencies
       run: |
-        python -m pip install --user --upgrade setuptools wheel
+        python -m pip install --user --upgrade setuptools wheel build
```
hugovk, line 19: `build` installs what it needs, can omit these:
```suggestion
        python -m pip install --user --upgrade build
```

pyproject.toml
hugovk, line 26: I think we can skip this wheel too:

```suggestion
requires = ["setuptools >= 35.0.2", "setuptools_scm >= 2.0.0, <3"]
```

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "pyperf"
+dynamic = ["version"]
```
CAM-Gerlach, line 31: Since you have a test to verify these remain in sync, you should consider continuing to decoratively specifying the version here, as a static version can make builds/dep resolution more efficient and deterministic, integrates better with some tooling and can avoid certain types of packaging problems.

```suggestion
version = "2.6.0"
```

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+license = {text = "MIT license"}
```
CAM-Gerlach, line 32:
```suggestion
license = {text = "MIT"}
```

PEP 639 is still in work, but in the meantime its worth still using standard SPDX identifiers here, and repeating "license" is rather redundant given this is the "license" field, after all

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+]
+dependencies = ["psutil>=5.9.0"]
+
+[project.entry-points.console_scripts]
+pyperf = "pyperf.__main__:main"
+
+[tool.setuptools.dynamic]
+version = {attr = "pyperf.__version__"}
+
+[tool.setuptools]
```
CAM-Gerlach, line 55:
```suggestion
[tool.setuptools]
```

If you specify the version statically above, this section can be eliminated. (It also avoids these sections being out of order.)

tox.ini
```diff
@@ -1,8 +1,9 @@
 [tox]
 envlist = py3, pypy3, doc, pep8
+isolated_build = True
 
 [testenv]
-deps=-r test-requirements.txt
+deps=importlib-metadata
```
hugovk, line 6: Can reuse the `dev` extras, also makes use of the conditional for `importlib-metadata`

```suggestion
extras=dev
```

I'm not sure if `tox` being in the extra will be a problem, is so, could have the `dev` extra with `tox` and a `test` extra with `importlib-metadata`

Or just remove `tox` from the `dev` extra

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
+dependencies = ["psutil>=5.9.0"]
+
+[project.entry-points.console_scripts]
```
CAM-Gerlach, line 49:
```suggestion
[project.scripts]
```

Fix this to use the correct config setting

.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install Tox and any other packages
-        run: pip install tox
+        run: pip install .[dev]
```
hugovk, line 33:
```suggestion
        run: pip install tox
```

Keep tox here, it installs all it needs in a virtual env

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
+dependencies = ["psutil>=5.9.0"]
```
CAM-Gerlach, line 47:
```suggestion
dependencies = ["psutil>=5.9.0"]

[project.optional-dependencies]
dev = [
    'tox',
    'importlib-metadata; python_version < "3.8"'
]
```

Move this here, to keep things in their logical order and avoid scattering the project table about (which is heavily discouraged in TOML without a very good reason)

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+version = {attr = "pyperf.__version__"}
+
+[tool.setuptools]
+packages = ["pyperf", "pyperf.tests"]
+
+[project.optional-dependencies]
+dev = [
+    'tox',
+    'importlib-metadata; python_version < "3.8"'
+]
```
CAM-Gerlach, line 62:
```suggestion
```

Moved this above

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+maintainers = [{name = "Dong-hee Na", email="donghee.na@python.org"}]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Topic :: Software Development :: Libraries :: Python Modules"
+]
```
CAM-Gerlach, line 46:
```suggestion
]
requires-python = ">=3.7"
```

Missing this; without it, pip and other installers will install `pyperf` on older Pythons it isn't compatible with, instead of falling back to a compatible version or raising an error. This also declares your version compatibility both for humans and machines.

CAM-Gerlach (commented):
Not 100% comprehensive, but this should fix the most important issues I noticed looking through it

.github/workflows/publish.yml
CAM-Gerlach: The only issue is that depending on the caching policy for isolated envs (which I can never recall precisely for different frontends), older versions that are installed by default may be used instead of the latest. I seem to remember this happening in the past, but maybe this isn't an issue nowadays...

pyproject.toml
CAM-Gerlach: Yes, though more important changes are required; see my suggestion

pyproject.toml
corona10: This was intended, I would like to reduce the defining multiple version semantics.
Currently, we have 3 sections about versioning.
- https://github.com/psf/pyperf/blob/d373c5e56c0257d2d7abd705b676bea25cf66566/setup.py#L27
- https://github.com/psf/pyperf/blob/d373c5e56c0257d2d7abd705b676bea25cf66566/pyperf/__init__.py#L3-L4
- https://github.com/psf/pyperf/blob/d373c5e56c0257d2d7abd705b676bea25cf66566/doc/conf.py#L58

Do you have better ideas?

tox.ini
```diff
@@ -1,8 +1,9 @@
 [tox]
 envlist = py3, pypy3, doc, pep8
+isolated_build = True
 
 [testenv]
-deps=-r test-requirements.txt
+deps=dev
```
CAM-Gerlach:
```suggestion
extras=dev
```

@hugovk 's suggestion was not applied correctly, which seems to be the cause of the build failure

CAM-Gerlach:
As a general tip that even experienced folks often don't remember or aren't aware of, you can easily apply some or all suggestions directly in one go instead of repeating the same changes yourself manually, which can be time-wasting, tedious and error prone (as was the case here), by going to the `Files changed` tab, clicking `Add to batch` on each suggestion, and then clicking `Commit` to commit them all in one go. If you want to modify a suggestion or make your own, you can also easily do so in your own reply comment and add/apply that alongside the others.

pyproject.toml
CAM-Gerlach: I understand; I just wanted to mention that there is some amount of cost to it. But there's nothing _wrong_ per-say about the current approach; if you're sure that's what you want, what you have would be the way to do it using base Setuptools.

I do suggest, though, moving the `tool.setuptools.dynamic` section below the `tool.setuptools` section to follow standard TOML conventions.

To avoid the duplication in the `conf.py`, since its a dynamic Python file, you can read the version from the `__init__.py` via any number of means—unfortunately, as it imports a bunch of stuff, best not to use `exec` or `importlib`, but you could extract it with `re`, `ast` or other methods.

A popular option for further single-sourcing both the version and other things is Setuptools_scm; the amount of single-sourcing is higher, but also the potential cost and complexities for some types of packaging scenarios. So, likely worth considering separately from this PR.

corona10:
@hugovk @CAM-Gerlach 

I apply most of your suggestions, would you like to take a look before merging this PR?

> suggestions directly in one go instead of repeating the same changes yourself manually

Actually, it was intended after I met some messy cases, but I would like to try using the batch command as possible.
Thanks for the suggestion.

CAM-Gerlach:
Review in work now...

> Actually, it was intended after I met some messy cases, but I would like to try using the batch command as possible.

Yeah, totally fine of course—I just know a lot of folks, even core devs, don't always know or remember this feature and as a reviewer I feel guilty when others, especially core devs like you, spend their valuable time implementing my suggestions. The other advantage for authors I didn't mention is that if something breaks with an applied suggestion, you know its the reviewer that messed up (as I all to often do :laughing: )

CAM-Gerlach (approved):
LGTM, thanks @corona10 :+1:

corona10:
Thank you @CAM-Gerlach @hugovk 
If there needs more improvement please let me know through the issue tracker :)

PR #151, opened 2022-11-30
pyproject.toml: Add Python3.6 as available Python version

corona10:
We dropped the support for Python 3.6 through https://github.com/psf/pyperf/pull/142.
Technically, it was available to be installed at Python 3.6 but since [pyperformance still uses Python 3.6,](https://github.com/python/pyperformance/pull/249#discussion_r1035191258) 
I think that we should sync the version support with the pyperformance team.

There was a restriction at `pyproject.toml` but it was never published to PyPI yet luckily.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,8 @@ jobs:
         python: ['3.11']
         include:
           - os: ubuntu-latest
+            python: '3.6'
+          - os: ubuntu-latest
             python: '3.7'
           - os: ubuntu-latest
             python: '3.8'
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,45 @@
+name: Check
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  pep8:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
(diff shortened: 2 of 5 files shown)


corona10:
cc @ericsnowcurrently

corona10:
## Option A
- Do not maintain Python 3.6 CI
- Downgrade requires-python into Python 3.6

## Option B
- Maintain requires-python as Python 3.7
- Drop support Python 3.6 at pyperformance too

## Option C
- Maintain Python 3.6 CI
- Downgrade requires-python into Python 3.6

vstinner:
The CI failed with:

> ERROR: Could not find a version that satisfies the requirement setuptools>=61

Is it possible to use older setuptools on Python <= 3.6?

pyproject.toml contains:

```
[build-system]
requires = ["setuptools >= 61"]
build-backend = "setuptools.build_meta"
```

Why is version 61 required?

corona10:
> Why is version 61 required?

I think that we can downgrade into 59

corona10:
@vstinner @ericsnowcurrently @mdboom

There looks some issue between the new package meta system with setuptools and Python 3.6
If we still need to support Python 3.6, I would like to suggest rollback the package build system into setup.py.

vstinner:
So pyperformance doesn't use Python 3.7 anymore?

corona10:
@vstinner we merged https://github.com/python/pyperformance/pull/249, so no support!

Issue #156, opened 2023-03-15
Support custom loop factory in bench_async_func


itamaro:
It may be desirable to use a custom event loop in an async benchmark (e.g. for benchmarking https://github.com/python/cpython/issues/97696)

`bench_async_func` currently doesn't support that - it will always use the default loop

PR #157, opened 2023-03-15
Support loop_factory parameter in bench_async_func

itamaro:
Fixes GH-156


methane (approved):

corona10 (commented):
Would you like to update the
https://github.com/psf/pyperf/blob/main/doc/changelog.rst
for this change as 2.6.0?

corona10 (approved):
lgtm

PR #131, opened 2022-05-23
Reject --tracemalloc in `pyperf command`

oraluben:
Close #130

PR #161, opened 2023-05-11
Fix tracemalloc error when action isn't defined

basepi:
Ref #131 

[We're using the `Runner`'s argument parsing](https://github.com/elastic/apm-agent-python-benchmarks/blob/1cacb6fef93ec570c15472c3df757eef1bfd2d28/run_bench.py#L36), but without using `pyperf` directly. This is resulting in an undefined `action`, which means that when we use the `--tracemalloc` argument, we get the following error:

```
  File ".../python3.8/site-packages/pyperf/_runner.py", line 334, in _process_args_impl

    if args.action == 'command':

AttributeError: 'Namespace' object has no attribute 'action'
```

This PR checks for the attribute before accessing it.

### DIFF
--- a/pyperf/_runner.py
+++ b/pyperf/_runner.py
@@ -331,7 +331,7 @@ class Runner:
             self._only_in_worker("--worker-task")
 
         if args.tracemalloc:
-            if args.action == 'command':
+            if hasattr(args, "action") and args.action == 'command':
                 raise CLIError('--tracemalloc cannot be used with pyperf command')
             try:
                 import tracemalloc   # noqa
--- a/pyperf/_runner.py
+++ b/pyperf/_runner.py
@@ -331,7 +331,7 @@ class Runner:
             self._only_in_worker("--worker-task")
 
         if args.tracemalloc:
-            if hasattr(args, "action") and args.action == 'command':
+            if getattr(args, 'action', None) == 'command':
                 raise CLIError('--tracemalloc cannot be used with pyperf command')
             try:
                 import tracemalloc   # noqa
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 2.6.1
+-------------
+
+* Fix a possible attribute error in argument parsing.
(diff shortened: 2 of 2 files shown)


vstinner:
Which command line did you use to trigger this error?

basepi:
We're triggering the runner with [our own script](https://github.com/elastic/apm-agent-python-benchmarks/blob/main/run_bench_commits.py). That script calls [another script](https://github.com/elastic/apm-agent-python-benchmarks/blob/main/run_bench.py) with the relevant `pyperf` args (such as `--tracemalloc`), which we parse [here](https://github.com/elastic/apm-agent-python-benchmarks/blob/1cacb6fef93ec570c15472c3df757eef1bfd2d28/run_bench.py#L36). The actual pyperf invocation we're using is [`runner.bench_func()`](https://github.com/elastic/apm-agent-python-benchmarks/blob/1cacb6fef93ec570c15472c3df757eef1bfd2d28/run_bench.py#L53).

It's an odd way of invoking pyperf. This is legacy code I've been slowly cleaning up and modernizing. Perhaps using `parse_args()` this way without actually using the `pyperf` command is not supported. But it worked until #131. That is the only place in the arg handling code where `args.action` is accessed, so upgrading `pyperf` resulted in an `AttributeError` for us.

pyperf/_runner.py
```diff
@@ -331,7 +331,7 @@ def _process_args_impl(self):
             self._only_in_worker("--worker-task")
 
         if args.tracemalloc:
-            if args.action == 'command':
+            if hasattr(args, "action") and args.action == 'command':
```
vstinner, line 334:
```suggestion
            if getattr(args, 'action', None) == 'command':
```

vstinner (commented):
Oh right, I can reproduce the issue with ``python3 doc/examples/bench_func.py --tracemalloc`` command.

basepi:
@vstinner Is there anything else pending here that I can help with?

basepi:
@vstinner Can we get this merged?

corona10 (approved):
lgtm

PR #163, opened 2023-06-03
.github: Use Trusted Publishers for deploy

corona10:


corona10:
cc @hugovk

hugovk:
Looks good.

Before merge, who can do the PyPI settings like in https://github.com/python/pyperformance/pull/300?

corona10:
> Before merge, who can do the PyPI settings like in

I did :)

hugovk (approved):

PR #171, opened 2023-11-02
pyperf: Prepare releasing 2.6.2

corona10:


doc/changelog.rst
```diff
@@ -1,8 +1,16 @@
 Changelog
 =========
 
+Version 2.6.2 (2023-11-02)
+---------------------------
+
+* Now, stats are collected during warmups.
```
vstinner, line 7:
```suggestion
* Now, pystats stats are collected during warmups.
```

vstinner (approved):
LGTM, but "stats" is too generic, I suggest adding "pystats".

PR #173, opened 2024-03-01
Fix mem_max_rss measurement on macOS

mdboom:
On BSD/darwin, ru_maxrss is defined in "bytes", unlike Linux where it is "kilobytes".  Therefore, this code was wrongly multiplying the result by 1024.  Rather than just fixing it here, it seems better to reuse the already correct `get_max_rss` function (used for calculating `command_max_rss`) elsewhere.


corona10 (approved):
LGTM nice catch :)

Issue #174, opened 2024-03-05
Disable pyperf for free-threaded build for a while


corona10:
See: https://github.com/python/cpython/issues/116024

vstinner:
Do you mean disable psutil?

PR #175, opened 2024-03-05
gh-174: Support free-threading CPython by disabling psutil related fe…

corona10:
…atures

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version 2.6.3 (2024-03-05)
+---------------------------
+
+* Support Free-threading CPython (PEP-703) by disabling psutil related features.
+  Patch by Donghee Na.
+* Fix mem_max_rss measurement on macOS.
+  Patch by Mike Droettboom.
+
 Version 2.6.2 (2023-11-02)
 ---------------------------
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,7 +5,7 @@ Version 2.6.3 (2024-03-05)
 ---------------------------
 
 * Support Free-threading CPython (PEP-703) by disabling psutil related features.
-  Relavant issue: https://github.com/python/cpython/issues/116024
+  Relevant issue: https://github.com/python/cpython/issues/116024.
   Patch by Donghee Na.
 * Fix mem_max_rss measurement on macOS.
   Patch by Mike Droettboom.
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -5,6 +5,7 @@ Version 2.6.3 (2024-03-05)
 ---------------------------
 
 * Support Free-threading CPython (PEP-703) by disabling psutil related features.
+  Relavant issue: https://github.com/python/cpython/issues/116024
   Patch by Donghee Na.
 * Fix mem_max_rss measurement on macOS.
   Patch by Mike Droettboom.
--- a/pyperf/_collect_metadata.py
+++ b/pyperf/_collect_metadata.py
@@ -11,8 +11,12 @@ except ImportError:
     resource = None
 
 try:
+    from pyperf._utils import IS_FREE_THREADING
     # Optional dependency
-    import psutil
+    if IS_FREE_THREADING:
+        psutil = None
+    else:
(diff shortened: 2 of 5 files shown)


corona10:
cc @colesbury

doc/changelog.rst
```diff
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+Version 2.6.3 (2024-03-05)
+---------------------------
+
+* Support Free-threading CPython (PEP-703) by disabling psutil related features.
```
vstinner, line 7: Please add a link to the issue.

pyperf/_collect_metadata.py
```diff
@@ -11,8 +11,12 @@
     resource = None
 
 try:
+    from pyperf._utils import IS_FREE_THREADING
     # Optional dependency
-    import psutil
+    if IS_FREE_THREADING:
```
vstinner, line 16: Please add a link to the issue to explain the skip. Same remark for similar changes.

vstinner (commented):
Can't we get psutil fixed instead of having such workaround?

corona10:
> Can't we get psutil fixed instead of having such workaround?

It's not related to psutil, it is more about packaging issue for a while.

doc/changelog.rst
corona10: Which link? PEP 703?

colesbury: The current blocker is https://github.com/pypa/packaging/issues/755, but there's ~3 more steps needed after that too (update packaging in pip, release pip, update pip in CPython).

vstinner: The issue describing why pyperf cannot use psutil on a Python Free Threading build.

corona10: Done.

doc/changelog.rst
```diff
@@ -5,6 +5,7 @@ Version 2.6.3 (2024-03-05)
 ---------------------------
 
 * Support Free-threading CPython (PEP-703) by disabling psutil related features.
+  Relavant issue: https://github.com/python/cpython/issues/116024
```
vstinner, line 8:
```suggestion
  Relevant issue: https://github.com/python/cpython/issues/116024.
```

pyperf/_cpu_utils.py
```diff
@@ -2,11 +2,16 @@
...
 
-from pyperf._utils import sysfs_path, proc_path, read_first_line
+from pyperf._utils import sysfs_path, proc_path, read_first_line, FREE_THREADING
 
 try:
     # Optional dependency
-    import psutil
+    # Currently there is a packaging issue for PEP-703,
+    # Until then psutil is disabled as a workaround.
+    if FREE_THREADING:
```
vstinner, line 11: Mmmh. Maybe instead of copying the same comment in many files, you can add a constant and reuse it, to have a single comment where the constant is defined. I'm making this comment since I would like to suggest to copy the URL to the bug.

vstinner (approved):
LGTM. I just have a last comment about the comments :-)

pyperf/_cpu_utils.py
corona10: Hmm, would you like to propose the suggestion more concretely?
I think that this PR will be reverted once the packaging issue is resolved :)

I though about get_psutil() function, do you think that it is better? But it will be a meaningless function once the issue is fixed.

pyperf/_utils.py
```diff
@@ -3,10 +3,11 @@
...
 import statistics
 import sys
+import sysconfig
 from shlex import quote as shell_quote   # noqa
 from shutil import which
 
-
+FREE_THREADING = bool(sysconfig.get_config_var('Py_GIL_DISABLED'))
 MS_WINDOWS = (sys.platform == 'win32')
 MAC_OS = (sys.platform == 'darwin')
```
vstinner, line 12: My suggestion is to add `USE_PSUTIL = FREE_THREADING` constant here and move comments explaining why Free Threading cannot use psutil right now, with the https://github.com/python/cpython/issues/116024 link.

corona10:
@vstinner Would you like to take a look the PR again?

vstinner (approved):
LGTM

PR #180, opened 2024-04-19
Include `CONFIG_ARGS` in Metadata

### LINKED ISSUE #179: Include `CONFIG_ARGS` in Metadata
It would be useful to have the information from the `CONFIG_ARGS` variable in the metadata for a benchmark file.

The use case is: determining whether two benchmarks were built from the same Python commit and built with the same options (specifically with regards to the new JIT and PGO).

For more context - I am experimenting with paralleling `pyperformance` across several machines, and for efficiency I am caching several builds of Python on each machine. In order to generate useful comparison data between many different benchmark files, it's useful to extract the commit, JIT and PGO status from the benchmark files themselves.

If there's a better way to determine that data, I'd be open to that.

I have a branch which adds in this info, if it would be useful to submit a PR.

JeffersGlass:
This PR includes the value of `sys.get_config_var('CONFIG_ARGS')` in the metadata for a benchmark run.

This value is _not_ added to `_bench._CHECKED_METADATA`, since it is often useful and desired to compare two benchmarks with different build configs.

Closes #179

### DIFF
--- a/pyperf/_collect_metadata.py
+++ b/pyperf/_collect_metadata.py
@@ -106,11 +106,11 @@ def collect_python_metadata(metadata):
         if cflags:
             cflags = normalize_text(cflags)
             metadata['python_cflags'] = cflags
-            
-        config_flags = sysconfig.get_config_var('CONFIG_ARGS')
-        if config_flags:
-            config_flags = normalize_text(config_flags)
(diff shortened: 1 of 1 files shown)


vstinner:
Merged, thank you.

PR #182, opened 2024-04-28
Minor optimizations by using comprehensions

cclauss:
% `ruff check --select=C4,PERF --ignore=PERF203`
```
pyperf/__main__.py:355:21: PERF401 Use a list comprehension to create a transformed list
pyperf/_bench.py:563:21: PERF401 Use a list comprehension to create a transformed list
pyperf/_compare.py:168:13: PERF401 Use a list comprehension to create a transformed list
pyperf/_compare.py:253:21: C414 Unnecessary `list` call within `sorted()`
pyperf/_compare.py:285:13: PERF401 Use a list comprehension to create a transformed list
pyperf/_compare.py:432:13: PERF401 Use a list comprehension to create a transformed list
pyperf/_cpu_utils.py:94:17: PERF402 Use `list` or `list.copy` to create a copy of a list
pyperf/_utils.py:133:21: PERF402 Use `list` or `list.copy` to create a copy of a list
pyperf/tests/test_bench.py:351:13: PERF401 Use a list comprehension to create a transformed list
Found 9 errors.
```
% `ruff rule PERF401`
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

### DIFF
--- a/pyperf/__main__.py
+++ b/pyperf/__main__.py
@@ -349,11 +349,7 @@ class Benchmarks:
     def group_by_name_ignored(self):
         names = set(self._group_by_name_names())
         for suite in self.suites:
-            ignored = []
-            for bench in suite:
-                if bench.get_name() not in names:
-                    ignored.append(bench)
-            if ignored:
+            if ignored := [bench for bench in suite if bench.get_name() not in names]:
                 yield (suite, ignored)
 
 
--- a/pyperf/_bench.py
+++ b/pyperf/_bench.py
@@ -557,10 +557,7 @@ class Benchmark(object):
         if include:
             old_runs = self._runs
             max_index = len(old_runs) - 1
-            runs = []
-            for index in only_runs:
-                if index <= max_index:
-                    runs.append(old_runs[index])
+            runs = [old_runs[index] for index in only_runs if index <= max_index]
         else:
             runs = self._runs
             max_index = len(runs) - 1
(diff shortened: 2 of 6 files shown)


corona10:
ditto: https://github.com/psf/pyperf/pull/183#issuecomment-2081878177

vstinner (commented):
I dislike this change. I don't think that these "micro optimizations" are worth it.

vstinner:
I suggest to ignore this recommendation instead of trying to address it just because ruff thinks that the code is slow.

PR #183, opened 2024-04-28
Add some ruff rules for pyupgrade

cclauss:
% `ruff check --select=SIM,UP --ignore=SIM105,SIM115,SIM117,UP015,UP031 --statistics | sort -k2`
```
23	SIM108	[*] Use ternary operator `char = '=' if level == 1 else '-'` instead of `if`-`else`-block
 1	SIM113	[ ] Use `enumerate()` for index variable `index` in `for` loop
 7	SIM114	[*] Combine `if` branches using logical `or` operator
10	UP004 	[*] Class `BaseTestCase` inherits from `object`
 1	UP009 	[*] UTF-8 encoding declaration is unnecessary
 1	UP020 	[*] Use builtin `open`
18	UP024 	[*] Replace aliased errors with `OSError`
 9	UP025 	[*] Remove unicode literals from strings
```
% `ruff rule UP024`
# os-error-alias (UP024)

Derived from the **pyupgrade** linter.

Fix is always available.

## What it does
Checks for uses of exceptions that alias `OSError`.

## Why is this bad?
`OSError` is the builtin error type used for exceptions that relate to the
operating system.

In Python 3.3, a variety of other exceptions, like `WindowsError` were
aliased to `OSError`. These aliases remain in place for compatibility with
older versions of Python, but may be removed in future versions.

Prefer using `OSError` directly, as it is more idiomatic and future-proof.

## Example
```python
raise IOError
```

Use instead:
```python
raise OSError
```

## References
- [Python documentation: `OSError`](https://docs.python.org/3/library/exceptions.html#OSError)

### DIFF
--- a/pyperf/_bench.py
+++ b/pyperf/_bench.py
@@ -86,7 +86,7 @@ def _cached_attr(func):
     return method
 
 
-class Run(object):
+class Run:
     # Run is immutable, so it can be shared/exchanged between two benchmarks
 
     __slots__ = ('_warmups', '_values', '_metadata')
@@ -319,7 +319,7 @@ class Run(object):
         return self._replace(metadata=metadata2)
 
 
-class Benchmark(object):
+class Benchmark:
     def __init__(self, runs):
         self._runs = []   # list of Run objects
         self._clear_runs_cache()
@@ -627,7 +627,7 @@ class Benchmark(object):
         self._replace_runs(new_runs)
 
 
-class BenchmarkSuite(object):
+class BenchmarkSuite:
     def __init__(self, benchmarks, filename=None):
         if not benchmarks:
             raise ValueError("benchmarks must be a non-empty "
@@ -727,7 +727,7 @@ class BenchmarkSuite(object):
         if isinstance(filename, bytes):
             suffix = b'.gz'
         else:
-            suffix = u'.gz'
+            suffix = '.gz'
 
         if filename.endswith(suffix):
             # Use lazy import to limit imports on 'import pyperf'
@@ -770,7 +770,7 @@ class BenchmarkSuite(object):
         if isinstance(filename, bytes):
             suffix = b'.gz'
         else:
-            suffix = u'.gz'
+            suffix = '.gz'
 
         if not replace and os.path.exists(filename):
             raise OSError(errno.EEXIST, "File already exists")
--- a/pyperf/_utils.py
+++ b/pyperf/_utils.py
@@ -155,7 +155,7 @@ def read_first_line(path, error=False):
         with open_text(path) as fp:
             line = fp.readline()
         return line.rstrip()
-    except IOError:
+    except OSError:
         if error:
             raise
         else:
@@ -282,7 +282,7 @@ def create_environ(inherit_environ, locale, copy_all):
     return env
 
 
-class _Pipe(object):
+class _Pipe:
     _OPEN_MODE = "r"
 
     def __init__(self, fd):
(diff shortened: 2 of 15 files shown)


corona10:
If you want to submit the patch-related lining tool, would you like to submit the patch, including the CI works?
cc @vstinner

pyperf/__main__.py
```diff
@@ -330,14 +330,8 @@ def group_by_name(self):
...
-                if show_name:
-                    if not show_filename:
-                        title = name
-                    else:
-                        # name is displayed in the group title
-                        title = filename
-                else:
-                    title = None
+                # name is displayed in the group title
+                title = (filename if show_filename else name) if show_name else None
```
vstinner, line 334: I'm not a fan of such change. I don't see how the existing code is bad/wrong.

pyperf/_bench.py
```diff
@@ -286,10 +283,7 @@ def _extract_metadata(self, name):
             raise KeyError("run has no metadata %r" % name)
 
         info = get_metadata_info(name)
-        if info.unit:
-            metadata = dict(self._metadata, unit=info.unit)
-        else:
-            metadata = None
+        metadata = dict(self._metadata, unit=info.unit) if info.unit else None
```
vstinner, line 286: For me, it's less readable than before.

vstinner (commented):
Would it be possible to exclude changes converting code to "expr1 if test else expr2"?

pyperf/_utils.py
```diff
@@ -193,9 +193,7 @@ def python_has_jit():
     implementation_name = python_implementation()
     if implementation_name == 'pypy':
         return sys.pypy_translation_info["translation.jit"]
-    elif implementation_name in ['graalpython', 'graalpy']:
-        return True
-    elif hasattr(sys, "pyston_version_info") or "pyston_lite" in sys.modules:
+    elif implementation_name in {'graalpython', 'graalpy'} or (hasattr(sys, "pyston_version_info") or "pyston_lite" in sys.modules):
```
vstinner, line 196: I dislike this line. It just makes the line longer and make complicated to read. I like to fit into 80 columns.

pyperf/tests/test_runner.py
```diff
@@ -249,9 +249,8 @@ def test_calibration_zero(self):
         def time_func(loops):
             return 0
 
-        with self.assertRaises(SystemExit):
-            with tests.capture_stdout() as stdout:
-                runner.bench_time_func('bench', time_func)
+        with self.assertRaises(SystemExit), tests.capture_stdout() as stdout:
```
vstinner, line 252: I prefer to write it on two lines for readability. Maybe just split the line on two lines using `\` line terminator.

vstinner (commented):
@corona10: What do you think of this change?

vstinner:
> @cclauss cclauss closed

This change combines many different changes. Some of them are uncontroversial, I have an opinion on the others :-)

cclauss:
These PRs are super easy to create locally with syntax like:
% `ruff check --select=UP004 --fix`

vstinner:
I'm fine with these changes, coming from the old time when pyperf worked on Python 2:

```
10	UP004 	[*] Class `BaseTestCase` inherits from `object`
 1	UP009 	[*] UTF-8 encoding declaration is unnecessary
 1	UP020 	[*] Use builtin `open`
18	UP024 	[*] Replace aliased errors with `OSError`
 9	UP025 	[*] Remove unicode literals from strings
```

vstinner (approved):
LGTM, thanks.

vstinner:
Merged, thanks.

PR #184, opened 2024-04-29
tox.ini: Replace flake8 with ruff

cclauss:
https://docs.astral.sh/ruff

As recommended at https://github.com/psf/pyperf/pull/183#issuecomment-2081878177

Should we rename the tox job from `pep8` to `lint` or not?


vstinner (approved):
LGTM.

@corona10: What do you think?

corona10 (approved):

PR #186, opened 2024-05-03
Fix logic in MSR availability test

mdboom:
As @vstinner [pointed out](https://github.com/psf/pyperf/pull/185#issuecomment-2092945317) the logic to turn off MSR setting on non-Intel hardware was reversed.


vstinner (approved):
LGTM

Issue #187, opened 2024-05-10
Benchmark with disabled GIL build and -Xgil=1 isn't enabling GIL


tonybaloney:
It seems that when you run `PYTHON_GIL=1 python benchmark.py` the benchmarked functions are being run with the GIL disabled and ignoring the flag.

<details><summary>Benchmark Code</summary>
<p>

```python
import pyperf
from multiprocessing import Process
from threading import Thread
try:
    import _xxsubinterpreters as interpreters
except ImportError:
    import _interpreters as interpreters

import itertools

DEFAULT_DIGITS = 2000
icount = itertools.count
islice = itertools.islice

def gen_x():
    return map(lambda k: (k, 4 * k + 2, 0, 2 * k + 1), icount(1))


def compose(a, b):
    aq, ar, as_, at = a
    bq, br, bs, bt = b
    return (aq * bq,
            aq * br + ar * bt,
            as_ * bq + at * bs,
            as_ * br + at * bt)


def extract(z, j):
    q, r, s, t = z
    return (q * j + r) // (s * j + t)


def gen_pi_digits():
    z = (1, 0, 0, 1)
    x = gen_x()
    while 1:
        y = extract(z, 3)
        while y != extract(z, 4):
            z = compose(z, next(x))
            y = extract(z, 3)
        z = compose((10, -10 * y, 0, 1), z)
        yield y


def calc_ndigits(n=DEFAULT_DIGITS):
    return list(islice(gen_pi_digits(), n))

test ="""
import itertools

DEFAULT_DIGITS = 2000
icount = itertools.count
islice = itertools.islice

def gen_x():
    return map(lambda k: (k, 4 * k + 2, 0, 2 * k + 1), icount(1))


def compose(a, b):
    aq, ar, as_, at = a
    bq, br, bs, bt = b
    return (aq * bq,
            aq * br + ar * bt,
            as_ * bq + at * bs,
            as_ * br + at * bt)


def extract(z, j):
    q, r, s, t = z
    return (q * j + r) // (s * j + t)


def gen_pi_digits():
    z = (1, 0, 0, 1)
    x = gen_x()
    while 1:
        y = extract(z, 3)
        while y != extract(z, 4):
            z = compose(z, next(x))
            y = extract(z, 3)
        z = compose((10, -10 * y, 0, 1), z)
        yield y


def calc_ndigits(n=DEFAULT_DIGITS):
    return list(islice(gen_pi_digits(), n))
calc_ndigits()
"""

[...]

PR #188, opened 2024-05-11
gh-187: Support environment variables PYTHON_GIL / PYTHON_CPU_COUNT

corona10:
* https://docs.python.org/3.13/using/cmdline.html#envvar-PYTHON_CPU_COUNT
* https://docs.python.org/3.13/using/cmdline.html#envvar-PYTHON_GIL

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,34 +22,54 @@ jobs:
       - run: tox -e ${{ matrix.tox-job }}
   build:
     runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} - ${{ matrix.python }} ${{ matrix.build }}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python: ['3.13-dev']
+        build: ['']
         include:
           - os: ubuntu-latest
             python: 'pypy3.10'
+            build: ''
           - os: ubuntu-latest
             python: 'pypy3.9'
+            build: ''
           - os: ubuntu-latest
             python: 'pypy3.8'
+            build: ''
           - os: ubuntu-latest
             python: '3.8'
+            build: ''
           - os: ubuntu-latest
             python: '3.9'
+            build: ''
           - os: ubuntu-latest
             python: '3.10'
+            build: ''
           - os: ubuntu-latest
             python: '3.11'
+            build: ''
           - os: ubuntu-latest
             python: '3.12'
+            build: ''
+          - os: ubuntu-latest
+            python: '3.13-dev'
+            build: 'free-threading'
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


corona10:
cc @tonybaloney

corona10:
@hugovk  PTAL for this PR too :)

hugovk (approved):

pyperf/_utils.py
```diff
@@ -266,8 +266,10 @@ def create_environ(inherit_environ, locale, copy_all):
     if copy_all:
         return os.environ
     env = {}
-
-    copy_env = ["PATH", "PYTHONPATH", "HOME", "TEMP", "COMSPEC", "SystemRoot", "SystemDrive"]
+    copy_env = ["PATH", "HOME", "TEMP", "COMSPEC", "SystemRoot", "SystemDrive"]
+    # TODO: In the future, maybe we should manage CPython environment variables
+    # depending on the Python version.
+    copy_env.extend(["PYTHONPATH", "PYTHON_CPU_COUNT", "PYTHON_GIL"])
```
vstinner, line 272: Just put it directly in the list, no?

pyperf/_utils.py
```diff
@@ -266,8 +266,10 @@ def create_environ(inherit_environ, locale, copy_all):
     if copy_all:
         return os.environ
     env = {}
-
-    copy_env = ["PATH", "PYTHONPATH", "HOME", "TEMP", "COMSPEC", "SystemRoot", "SystemDrive"]
+    copy_env = ["PATH", "HOME", "TEMP", "COMSPEC", "SystemRoot", "SystemDrive"]
+    # TODO: In the future, maybe we should manage CPython environment variables
```
vstinner, line 270: There is no need to check the Python version. You can remove the TODO.

pyperf/_utils.py
corona10: Well let's separate CPython specific variables :)

vstinner: You can separate them with a comment and/or a newline.

corona10: Well, I am not sure why you want to maintain the single list declaration even if this is not a performance-sensitive code

.github/workflows/build.yml
```diff
@@ -22,32 +22,54 @@ jobs:
       - run: tox -e ${{ matrix.tox-job }}
   build:
     runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }} - ${{ matrix.python }} ${{ matrix.build }}
```
vstinner, line 25: The PR title is only about "env var", but these changes look unrelated to env vars. You might merge them in a separated PR.

vstinner:
Remark: If clearing env vars is a surprising/bad default behavior for end users, we can also just remove it? I did it to make benchmarks easier to reproduce.

corona10:
>Remark: If clearing env vars is a surprising/bad default behavior for end users, we can also just remove it? I did it to make benchmarks easier to reproduce.

I think that the propagation is the more precise way. I prefer to support it as same as currently.

.github/workflows/build.yml
corona10: Done

vstinner (approved):
LGTM

PR #192, opened 2024-06-12
Fix the max_mem_rss measurement

mdboom:
In https://github.com/psf/pyperf/commit/c9604435bc46273f3b6e51101ea4091d57a2020b, which fixed a bug in scaling the `mem_max_rss` value on Darwin, I inadvertently changed the measurement from using `RUSAGE_SELF` to `RUSAGE_CHILDREN`.  This has resulted in the measurements being much coarser and incorrect on Darwin.  This reverts to the old correct behavior.

### DIFF
--- a/pyperf/_process_time.py
+++ b/pyperf/_process_time.py
@@ -52,6 +52,11 @@ def merge_profile_stats_files(src, dst):
 
 
 def bench_process(loops, args, kw, profile_filename=None):
+    if resource is not None:
+        rusage_children = resource.RUSAGE_CHILDREN
+    else:
+        rusage_children = 0
+
     max_rss = 0
     range_it = range(loops)
     start_time = time.perf_counter()
@@ -61,7 +66,7 @@ def bench_process(loops, args, kw, profile_filename=None):
         args = [args[0], "-m", "cProfile", "-o", temp_profile_filename] + args[1:]
 
     for _ in range_it:
-        start_rss = get_max_rss(resource.RUSAGE_CHILDREN)
+        start_rss = get_max_rss(rusage_children)
 
         proc = subprocess.Popen(args, **kw)
         with proc:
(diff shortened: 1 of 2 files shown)


pyperf/_process_time.py
```diff
@@ -26,9 +26,11 @@
     resource = None
 
 
-def get_max_rss():
+def get_max_rss(resource_type=None):
     if resource is not None:
-        usage = resource.getrusage(resource.RUSAGE_CHILDREN)
+        if resource_type is None:
+            resource_type = resource.RUSAGE_CHILDREN
```
vstinner, line 32: We should not use RUSAGE_SELF on Linux, FreeBSD and other platforms?

mdboom: This function is called in two contexts -- one when the benchmark runs entirely in a separate subprocess, in which case we want `CHILDREN`, and one when in which it runs in the same process, where we want `SELF`.  Saying that, it would probably clearer to not have a default value for `resource_type` here.

pyperf/_process_time.py
```diff
@@ -61,7 +61,7 @@ def bench_process(loops, args, kw, profile_filename=None):
         args = [args[0], "-m", "cProfile", "-o", temp_profile_filename] + args[1:]
 
     for _ in range_it:
-        start_rss = get_max_rss()
+        start_rss = get_max_rss(resource.RUSAGE_CHILDREN)
```
vstinner, line 64: `resource.RUSAGE_CHILDREN` doesn't work on Windows: resource module is not available there.

You should add an abstraction, like having a children=True parameter for get_max_rss().

pyperf/_process_time.py
```diff
@@ -26,9 +26,13 @@
     resource = None
 
 
-def get_max_rss():
+def get_max_rss(children=False):
```
vstinner, line 29: You may want to make the parameter mandatory:

```suggestion
def get_max_rss(*, children):
```

vstinner (approved):
LGTM

PR #193, opened 2024-06-24
Add hooks for calling arbitrary code around each benchmark

mdboom:
This change adds a concept of "hooks" to pyperf for running arbitrary code before and after the actual internal benchmarking code.  These are implemented with entry points, so it should be possible to write an external plugin to add new integrations and functionality.

There are a few real use cases that the Faster CPython team already needs this for:

- Integration with [codspeed](codspeed.io) to turn on/off the instrumentation only around the benchmarking code
- Doing coverage analysis of benchmarks
- Adding `sys.monitoring` hooks around benchmarks to do custom analysis of which specific bytecodes are run during benchmarking

The `pystats` integration that already exists in pyperf also should have, in hindsight, been implemented this way, so this PR also refactors things to do that, and it serves as a proof-of-concept of the new hook functionality.

~**This PR needs tests** but I thought I would submit a draft first to ensure this seems like the right direction first.~


vstinner (commented):
I like the concept of hooks.

Would it be possible to use importlib.metadata instead of setuptools pkg_resources? See: https://docs.python.org/dev/library/importlib.metadata.html#entry-points

mdboom:
> Would it be possible to use importlib.metadata instead of setuptools pkg_resources? See: https://docs.python.org/dev/library/importlib.metadata.html#entry-points

Sure -- I've never used entry points before -- I hadn't realized that's the new canonical way.

vstinner (approved):
LGTM overall. I don't know entry points, so I cannot review this change in details.

ericsnowcurrently (approved):
LGTM

corona10 (approved):

PR #194, opened 2024-07-16
Allow setting the amount of warmup values in Runner

PiispaH:
This PR enables the user to set the amount of warmup values in Runner once again.


mdboom (approved):
LGTM

PR #197, opened 2024-07-29
Make hooks work in bench_command subprocesses

mdboom:
This is a follow-on to #193 that makes hooks also work inside of subprocesses created by `bench_command`.

### DIFF
--- a/pyperf/_command.py
+++ b/pyperf/_command.py
@@ -1,4 +1,5 @@
 import functools
+import json
 import os.path
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


corona10 (commented):
I need to understand the detail :)
Please give me enough time by this week.

pyperf/_process_time.py
```diff
@@ -125,15 +151,21 @@ def main():
...
+        dt, max_rss = bench_process(loops, args, kw, profile_filename)
 
     if devnull is not None:
         devnull.close()
 
+    for hook in hook_managers.values():
+        hook.teardown(metadata)
+
     # Write timing in seconds into stdout
     print(dt)
```
corona10, line 166: @mdboom Would you like to extract printing data logic into a single function and then add a comment about the function from the parsing metadata logic at bench_command to explain why we parse that way?

pyperf/_command.py
```diff
@@ -23,18 +24,22 @@ def bench_command(command, task, loops):
                         % proc.returncode)
 
     rss = None
+    metadata = {}
     try:
         lines = output.splitlines()
         timing = float(lines[0])
-        if len(lines) >= 2:
-            rss = int(lines[1])
+        rss = int(lines[1])
```
corona10, line 31: Please explain about the parsing logic.
see: https://github.com/psf/pyperf/pull/197/files#r1718097565 :)

corona10 (changes requested):
Overall looks good I left a nit comment.

corona10 (approved):
let's just merge

PR #199, opened 2024-08-09
Use KiB and MiB as units instead of kB and MB.

### LINKED ISSUE #198: Minor issue regarding memory units (kB vs KiB)
The units in the output when `track-memory` is enabled say `kB` and `MB`. However, as far as I can tell, the correct units are KiB and MiB? It's a minor issue, and might not be worth changing due to the hassle of changing the output for downstream applications, but I found it confusing at first that the summary didn't match up with the actual numbers in the JSON file.

I wasn't sure if it was worth opening an issue for such a minor issue, but I thought it might be worth making a concious decision on whether to use precise SI units.

Sorry if there's already been a discussion on this, I tried looking but I couldn't find anything.

yngvem:
Previously, pyperf stated the memory usage as kB and MB, while strictly speaking, it was measured in KiB and MiB.
This commit makes it clear that the memory is measured using 2^10 and 2^20 bytes, not 10^3 and 10^6.

Closes #198 

Example output:
```raw
$ python3 -m pyperf timeit '[1,2]*1000' --track-memory
.....................
Mean +- std dev: 10.2 MiB +- 78.2 kiB
```

### DIFF
--- a/pyperf/_formatter.py
+++ b/pyperf/_formatter.py
@@ -30,9 +30,9 @@ def format_filesize(size):
             return '%.0f byte' % size
 
     if size > 10 * 1024 * 1024:
-        return '%.1f MB' % (size / (1024.0 * 1024.0))
+        return '%.1f MiB' % (size / (1024.0 * 1024.0))
 
-    return '%.1f kB' % (size / 1024.0)
+    return '%.1f kiB' % (size / 1024.0)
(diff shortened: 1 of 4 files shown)


vstinner (approved):
LGTM

corona10 (approved):

PR #202, opened 2024-09-06
Fix deprecated method in bench_time_func example

hajimes:
Hi,
The example code in `bench_time_func.py` still uses `pyperf.perf_counter()`, which was deprecated in version 2.0.0. This pull request updates it to use `time.perf_counter()`, as recommended by the API documentation.


vstinner (approved):
LGTM

vstinner:
Merged, thanks!

PR #204, opened 2024-09-16
Fix "std dev" link

vstinner:


fzakaria:
Does it help just to emit the standard deviation?

I was reading that knowing the standard deviation for a sample set isn't very helpful; I think it gives off the wrong impression that it's used as a confidence interval.

Reading the link - my feeling was the intent was for it to be a 95% confidence interval especially since the t-test value was calcualte.

vstinner:
The PR fix the documentation of the current implementation: describe correctly what's being used.

fzakaria:
sorry i wasn't clear; the output of `stats` has a +- on the standard deviation; that gives the impression like it's the confidence interval no?

vstinner:
cc @corona10

corona10:
I will take a look during the sprint.

corona10 (approved):
lgtm

PR #205, opened 2024-09-24
Implement --timeout when running benchmarks

diegorusso:
If the benchmark execution is exceeding the timeout execution, pyperf exits with an error 124. This error can be caught by pyperformance or other tool and report it back to the user.

### DIFF
--- a/pyperf/_manager.py
+++ b/pyperf/_manager.py
@@ -7,6 +7,8 @@ from pyperf._formatter import format_number
 from pyperf._utils import MS_WINDOWS, create_environ, create_pipe, popen_killer
 
 
+EXIT_TIMEOUT = 60
+
 # Limit to 5 calibration processes
 # (10 if calibration is needed for loops and warmups)
 MAX_CALIBRATION = 5
@@ -107,7 +109,7 @@ class Manager:
             with popen_killer(proc):
                 try:
                     bench_json = rpipe.read_text()
-                    exitcode = proc.wait()
+                    exitcode = proc.wait(timeout=EXIT_TIMEOUT)
                 except TimeoutError as exc:
                     print(exc)
                     sys.exit(124)
--- a/pyperf/_manager.py
+++ b/pyperf/_manager.py
@@ -69,6 +69,9 @@ class Manager:
         if args.profile:
             cmd.extend(['--profile', args.profile])
 
+        if args.timeout:
+            cmd.extend(['--timeout', str(args.timeout)])
+
         if args.hook:
             for hook in args.hook:
                 cmd.extend(['--hook', hook])
@@ -83,7 +86,7 @@ class Manager:
                              self.args.locale,
                              self.args.copy_env)
 
-        rpipe, wpipe = create_pipe()
+        rpipe, wpipe = create_pipe(timeout=self.args.timeout)
         with rpipe:
             with wpipe:
                 warg = wpipe.to_subprocess()
@@ -102,10 +105,12 @@ class Manager:
                 proc = subprocess.Popen(cmd, env=env, **kw)
 
             with popen_killer(proc):
-                with rpipe.open_text() as rfile:
-                    bench_json = rfile.read()
-
-                exitcode = proc.wait()
+                try:
+                    bench_json = rpipe.read_text()
+                    exitcode = proc.wait()
+                except TimeoutError as exc:
+                    print(exc)
+                    sys.exit(124)
 
         if exitcode:
             raise RuntimeError("%s failed with exit code %s"
--- a/pyperf/_manager.py
+++ b/pyperf/_manager.py
@@ -88,7 +88,7 @@ class Manager:
                              self.args.locale,
                              self.args.copy_env)
 
-        rpipe, wpipe = create_pipe(timeout=self.args.timeout)
+        rpipe, wpipe = create_pipe()
         with rpipe:
             with wpipe:
                 warg = wpipe.to_subprocess()
@@ -108,7 +108,7 @@ class Manager:
 
             with popen_killer(proc):
                 try:
-                    bench_json = rpipe.read_text()
+                    bench_json = rpipe.read_text(timeout=self.args.timeout)
                     exitcode = proc.wait(timeout=EXIT_TIMEOUT)
                 except TimeoutError as exc:
                     print(exc)
--- a/pyperf/_utils.py
+++ b/pyperf/_utils.py
@@ -1,9 +1,11 @@
 import contextlib
 import math
 import os
+import select
 import statistics
 import sys
 import sysconfig
+import time
 from shlex import quote as shell_quote   # noqa
 from shutil import which
 
@@ -286,8 +288,9 @@ def create_environ(inherit_environ, locale, copy_all):
 class _Pipe:
     _OPEN_MODE = "r"
 
-    def __init__(self, fd):
+    def __init__(self, fd, timeout=None):
         self._fd = fd
+        self._timeout = timeout
         self._file = None
         if MS_WINDOWS:
             self._handle = msvcrt.get_osfhandle(fd)
@@ -317,8 +320,33 @@ class _Pipe:
 class ReadPipe(_Pipe):
     def open_text(self):
         file = open(self._fd, "r", encoding="utf8")
+        if self._timeout:
+            os.set_blocking(file.fileno(), False)
         self._file = file
         return file
+
+    def read_text(self):
+        with self.open_text() as rfile:
+            if self._timeout:
+              return self._read_text_timeout(rfile, self._timeout)
+            else:
+                return rfile.read()
+
+    def _read_text_timeout(self, rfile, timeout):
+        start_time = time.time()
+        while True:
+            if time.time() - start_time > timeout:
+                raise TimeoutError(f"Timed out after {timeout} seconds")
+            ready, _, _ = select.select([rfile], [], [], timeout)
+            if ready:
+                data = rfile.read()
+                if data:
+                    return data
+                else:
+                    break
+            else:
+                pass
+
 
 
 class WritePipe(_Pipe):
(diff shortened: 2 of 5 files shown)


pyperf/_utils.py
```diff
@@ -317,9 +320,34 @@ def __exit__(self, *args):
...
+            os.set_blocking(file.fileno(), False)
         self._file = file
         return file
 
+    def read_text(self):
+        with self.open_text() as rfile:
+            if self._timeout:
+              return self._read_text_timeout(rfile, self._timeout)
+            else:
+                return rfile.read()
```
vstinner, line 333:
```suggestion
            if self._timeout is not None:
                return self._read_text_timeout(rfile, self._timeout)
            else:
                return rfile.read()
```

pyperf/_utils.py
```diff
@@ -317,9 +320,34 @@ def __exit__(self, *args):
...
+
+    def _read_text_timeout(self, rfile, timeout):
+        start_time = time.time()
```
vstinner, line 336: Can you please use time.monotonic() instead? It's not affected by DST or sys admin changing manually the system clock.

pyperf/_utils.py
```diff
@@ -317,9 +320,34 @@ def __exit__(self, *args):
...
+        while True:
+            if time.time() - start_time > timeout:
+                raise TimeoutError(f"Timed out after {timeout} seconds")
+            ready, _, _ = select.select([rfile], [], [], timeout)
+            if ready:
```
vstinner, line 341: To avoid one indentation level:

```
if not ready:
    continue
data = ...
```

pyperf/_manager.py
```diff
@@ -102,10 +105,12 @@ def spawn_worker(self, calibrate_loops, calibrate_warmups):
                 proc = subprocess.Popen(cmd, env=env, **kw)
 
             with popen_killer(proc):
-                with rpipe.open_text() as rfile:
-                    bench_json = rfile.read()
-
-                exitcode = proc.wait()
+                try:
+                    bench_json = rpipe.read_text()
+                    exitcode = proc.wait()
```
vstinner, line 110: Can you add an EXIT_TIMEOUT=60 constant (at the module level) and use it for `wait(timeout=EXIT_TIMEOUT)`? A process *should* exit immediately once the pipe is closed, but it's better to be safe here.

pyperf/_utils.py
```diff
@@ -317,9 +320,34 @@ def __exit__(self, *args):
...
+                data = rfile.read()
+                if data:
+                    return data
```
vstinner, line 344: You should read a few characters (ex: 1024) and put it in a list, rather than blocking on `.read()`, no?

Once done (once data is empty), `return ''.join(output)`.

vstinner (commented):
I'm fine with the overall approach, but here is a review on the actual implementation.

pyperf/_utils.py
diegorusso: The data that we read from these pipes are not massive so it should not be a big problem but I agree with you, let's break it down in smaller chunks.

vstinner: select() doesn't tell you how many bytes are available. Only that at least one byte is available.

diegorusso: I've implemented, have a look!

pyperf/_utils.py
```diff
@@ -317,9 +320,33 @@ def __exit__(self, *args):
 class ReadPipe(_Pipe):
     def open_text(self):
         file = open(self._fd, "r", encoding="utf8")
+        if self._timeout:
+            os.set_blocking(file.fileno(), False)
         self._file = file
         return file
 
+    def read_text(self):
```
vstinner, line 328: Can you move the timeout parameter from `__init__()` to this method instead?

It's a little bit strange to pass a timeout to the constructor whereas it creates two file descriptors (read end and write end).

pyperf/_utils.py
```diff
@@ -317,9 +320,33 @@ def __exit__(self, *args):
...
+        with self.open_text() as rfile:
+            if self._timeout is not None:
+                return self._read_text_timeout(rfile, self._timeout)
+            else:
+                return rfile.read()
```
vstinner, line 333: I don't think that io.TextIOWrapper, created by open_text(), handles well non-blocking file, and so I would prefer to work on "raw" file descriptors:

```suggestion
        if self._timeout is not None:
            return self._read_text_timeout(self._timeout)
        else:
            with self.open_text() as rfile:
                return rfile.read()
```

pyperf/_utils.py
```diff
@@ -317,9 +320,33 @@ def __exit__(self, *args):
...
+            if time.monotonic() - start_time > timeout:
+                raise TimeoutError(f"Timed out after {timeout} seconds")
+            ready, _, _ = select.select([rfile], [], [], timeout)
+            if not ready:
+                continue
+            data = rfile.read(1024)
+            if not data:
+                break
+            output.append(data)
+        return "".join(output)
```
vstinner, line 348: Something like:

```suggestion
        fd = self.fd()
        os.set_blocking(fd, False)

        start_time = time.monotonic()
        output = []
        while True:
            if time.monotonic() - start_time > timeout:
                raise TimeoutError(f"Timed out after {timeout} seconds")
            ready, _, _ = select.select([fd], [], [], timeout)
            if not ready:
                continue
            try:
                data = os.read(fd, 1024)
            except BlockingIOError:
                continue
            if not data:
                break
            output.append(data)

        data = b''.join(output)
        return data.decode('utf8')
```

pyperf/tests/test_runner.py
```diff
@@ -149,6 +149,24 @@ def test_pipe(self):
...
+    def test_pipe_with_timeout(self):
+        rpipe, wpipe = create_pipe(timeout=0.1)
+        with rpipe:
+            with wpipe:
+                arg = wpipe.to_subprocess()
+                # Don't close the file descriptor, it is closed by
+                # the Runner class
+                wpipe._fd = None
+
+                self.exec_runner('--pipe', str(arg), '--worker', '-l1', '-w1')
```
vstinner, line 161: Is this code really useful? It can maybe just be removed no?

We don't need wpipe, so you can just call: wpipe.close().

diegorusso: This is needed to write some "real" data to the wpipe so that we can read it back.

pyperf/_utils.py
diegorusso: Implemented. Have a look please.

pyperf/tests/test_runner.py
diegorusso: it will produce something like this
```
{"benchmarks":[{"runs":[{"values":[1.0,1.0,1.0],"warmups":[[1,1.0]]}]}],"metadata":{"aslr":"Full randomization","boot_time":"2024-09-21 15:48:34","cpu_config":"idle:none","cpu_count":8,"date":"2024-09-25 18:02:23.847103","duration":0.007623481971677393,"hostname":"ubuntu-vm","load_avg_1min":0.28,"loops":1,"mem_max_rss":27525120,"name":"bench","perf_version":"2.7.0","platform":"Linux-6.8.0-44-generic-aarch64-with-glibc2.39","python_cflags":"-fno-strict-overflow -Wsign-compare -DNDEBUG -g -O2 -Wall","python_compiler":"GCC 13.2.0","python_config_args":"'--enable-shared' '--prefix=/usr' '--libdir=/usr/lib/aarch64-linux-gnu' '--enable-ipv6' '--enable-loadable-sqlite-extensions' '--with-dbmliborder=bdb:gdbm' '--with-computed-gotos' '--without-ensurepip' '--with-system-expat' '--with-dtrace' '--with-ssl-default-suites=openssl' '--with-wheel-pkg-dir=/usr/share/python-wheels/' 'MKDIR_P=/bin/mkdir -p' 'CC=aarch64-linux-gnu-gcc'","python_executable":"/home/user/work/ce-sw/repos/pyperformance/venv/bin/python","python_implementation":"cpython","python_version":"3.12.3 (64-bit)","runnable_threads":1,"timer":"fake_clock, resolution: 1.00 sec","unit":"second","uptime":353629.8471829891},"version":"1.0"}
```

pyperf/_utils.py
diegorusso: Done.

vstinner (commented):
Nice, the change works as expected :-)

```
$ python -m pyperf timeit 'import time; time.sleep(5)' --timeout=2
Timed out after 2 seconds
```

vstinner (approved):
LGTM

pyperf/tests/test_runner.py
```diff
@@ -149,6 +152,33 @@ def test_pipe(self):
...
+                    rpipe.read_text(timeout=0.1)
+                self.assertEqual(str(cm.exception),
+                         'Timed out after 0.1 seconds')
+
+            # Mock the select to make the read pipeline ready
+            with mock.patch('pyperf._utils.select.select',
+                            return_value=(True, False, False)):
+                bench_json = rpipe.read_text(timeout=0.1)
+                self.assertEqual(bench_json,
+                                tests.benchmark_as_json(result.bench))
```
vstinner, line 180: The test fails on Windows because of different newline:

```
'(...)"version":"1.0"}\r\n' != '(...)"version":"1.0"}\n'
```

You can strip newlines in the comparison:

```suggestion
                self.assertEqual(bench_json.rstrip(),
                                 tests.benchmark_as_json(result.bench).rstrip())
```

diegorusso: Done.

pyperf/tests/test_runner.py
```diff
@@ -176,8 +173,8 @@ def test_pipe_with_timeout(self):
             with mock.patch('pyperf._utils.select.select',
                             return_value=(True, False, False)):
                 bench_json = rpipe.read_text(timeout=0.1)
-                self.assertEqual(bench_json,
-                                tests.benchmark_as_json(result.bench))
+                self.assertEqual(bench_json.rstrip(),
+                                tests.benchmark_as_json(result.bench).rstrip())
```
vstinner, line 177:
```suggestion
                                 tests.benchmark_as_json(result.bench).rstrip())
```

vstinner:
Merged, thanks.

PR #210, opened 2024-12-17
Add warnings about too few or too many samples

mdboom:
Related to https://github.com/python/pyperformance/issues/372.  Once this is merged, we can reduce the number of iterations in pyperformance.

### DIFF
--- a/pyperf/_bench.py
+++ b/pyperf/_bench.py
@@ -424,6 +424,35 @@ class Benchmark:
             raise ValueError("MAD must be >= 0")
         return value
 
+    def required_nsamples(self):
+        """
+        Determines the number of samples that would be required to have 95%
+        certainty that the samples have a variance of less than 1%.
+
+        This is described in this Wikipedia article about estimating the sampling of
+        a mean:
+
+        https://en.wikipedia.org/wiki/Sample_size_determination#Estimation_of_a_mean
+        """
+        # Get the means of the values per run
+        values = []
+        for run in self._runs:
+            values.append(statistics.mean(run.values))
+
+        total = math.fsum(values)
+        mean = total / len(values)
+        stddev = statistics.stdev(values)
+        # Normalize the stddev so we can target "percentage changed" rather than
+        # absolute time
+        sigma = stddev / mean
+
+        # 95% certainty
+        Z = 1.96
+        # 1% variation
+        W = 0.01
+
+        return int(math.ceil((4 * Z ** 2 * sigma ** 2) / (W ** 2)))
+
     def percentile(self, p):
         if not (0 <= p <= 100):
             raise ValueError("p must be in the range [0; 100]")
--- a/pyperf/_bench.py
+++ b/pyperf/_bench.py
@@ -424,17 +424,23 @@ class Benchmark:
             raise ValueError("MAD must be >= 0")
         return value
 
-    def required_nsamples(self):
+    def required_nprocesses(self):
         """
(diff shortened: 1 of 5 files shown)


pyperf/_bench.py
```diff
@@ -424,6 +424,39 @@ def median_abs_dev(self):
...
+        """
+        Determines the number of samples that would be required to have 95%
+        certainty that the samples have a variance of less than 1%.
+
+        This is described in this Wikipedia article about estimating the sampling of
+        a mean:
+
+        https://en.wikipedia.org/wiki/Sample_size_determination#Estimation_of_a_mean
+        """
+        # Get the means of the values per run
```
vstinner, line 437: Why not computing the mean only once, for all values of all runs?

pyperf/_cli.py
```diff
@@ -457,6 +466,16 @@ def format_checks(bench, lines=None):
...
 
+    if (
+        required_nsamples is not None and
+        required_nsamples < len(bench._runs) * 0.75
+    ):
+        lines.append("Benchmark was run more times than necessary to get a stable result.")
+        lines.append(
+            "Consider passing processes=%d to the Runner constructor to save time." %
+            required_nsamples
+        )
```
vstinner, line 477: This warning may be a little bit annoying. Maybe only show it in the "pyperf check" command? https://pyperf.readthedocs.io/en/latest/cli.html#check-cmd

pyperf/_bench.py
```diff
@@ -424,6 +424,39 @@ def median_abs_dev(self):
             raise ValueError("MAD must be >= 0")
         return value
 
+    def required_nsamples(self):
```
vstinner, line 427: If you want to add a public function, please document it at: https://pyperf.readthedocs.io/en/latest/api.html#benchmark-class

pyperf/_bench.py
mdboom: Because for some benchmarks, cache effects are visible within the same process.  For example, pylint takes about 30% longer during the first iteration than the subsequent 2 iterations.  One could argue that's a bad benchmark, but it's common enough that we should control for it.  There's some more discussion here: https://github.com/faster-cpython/bench_runner/issues/318#issuecomment-2537260960

That said, it's definitely worth putting a comment about that here.

pyperf/_bench.py
mdboom: Good point.  I think we do want it to public (for the same reason the other statistics methods are public).

pyperf/_bench.py
```diff
@@ -455,6 +462,7 @@ def required_nsamples(self):
         # 1% variation
         W = 0.01
 
+        # (4Z²σ²)/(W²)
         return int(math.ceil((4 * Z ** 2 * sigma ** 2) / (W ** 2)))
```
vstinner, line 466:
```suggestion
        return math.ceil((4 * Z ** 2 * sigma ** 2) / (W ** 2))
```

pyperf/tests/test_perf_cli.py
```diff
@@ -635,6 +628,14 @@ def test_slowest(self):
 
     def test_check_stable(self):
         stdout = self.run_command('check', TELCO)
+        self.assertTrue(
```
vstinner, line 631: I suggest using assertIn() instead.

vstinner (approved):
The code change LGTM. I didn't check the maths behind required_nprocesses().

PR #212, opened 2025-01-18
.github: Add ARM64 CI for pyperf

corona10:
See: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
     name: ${{ matrix.os }} - ${{ matrix.python }} ${{ matrix.build }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


corona10:
cc @diegorusso

vstinner:
Is it useful to test a different architecture? pyperf is mostly platform agnostic. But well, why not.

vstinner (approved):

corona10:
If we need ARM specific feature it would be helpful :)

diegorusso:
Thanks for the PR. I think it is always useful to validate the code on different platforms. Also pyperf depends on `psutil` which comes as binary distribution. It is good to check that the actual installation of psutil happens without any problem on all platforms (even though this  should be tested by them).

PR #218, opened 2025-08-09
ReadTheDocs: use latest Ubuntu and Python

vstinner:


vstinner:
See the docs:
* https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-tools-python
* https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os

vstinner:
cc @corona10

corona10 (approved):
lgtm

Issue #166, opened 2023-07-27
bench_func returning None for all but the first benchmark


axman6:
I'm trying to write a benchmark suite for several functions and use BenchmarkSuite to write the results to a json file. I have the following code (irrelevant parts removed):

```python
import pyperf
import decimation_accel as dec
import decimation_accel_opt as opt


results_file = 'decimation.json'

runner = pyperf.Runner()


def main() -> None:

    small = alloc_args(1024)
    small10 = alloc_args(1024, rate=10)
    small100 = alloc_args(1024, rate=100)
    medium = alloc_args(1024 * 10)
    benchmarks: List[pyperf.Benchmark] = []


    benchmarks.append(runner.bench_func('dec.decimate_minmax(small) all', dec.minmax, *small))
    benchmarks.append(runner.bench_func('opt.decimate_minmax(small) all', opt.minmax, *small))
    benchmarks.append(runner.bench_func('dec.decimate_minmax(small10) all', dec.minmax, *small10))
    benchmarks.append(runner.bench_func('opt.decimate_minmax(small10) all', opt.minmax, *small10))
    benchmarks.append(runner.bench_func('dec.decimate_minmax(small100) all', dec.minmax, *small100))
    benchmarks.append(runner.bench_func('opt.decimate_minmax(small100) all', opt.minmax, *small100))
    benchmarks.append(runner.bench_func('dec.decimate_minmax(medium) all', dec.minmax, *medium))
    benchmarks.append(runner.bench_func('opt.decimate_minmax(medium) all', opt.minmax, *medium))

    print(benchmarks)

    suite = pyperf.BenchmarkSuite(benchmarks)  # filename='decimation.json')

    suite.dump(results_file)


if __name__ == "__main__":
    main()
```
 However if I run this file, I get the following output:

```
[<Benchmark 'dec.decimate_minmax(small) all' with 1 runs>, None, None, None, None, None, None, None]
Traceback (most recent call last):
  File "benchmarks.py", line 60, in <module>
    main()
  File "benchmarks.py", line 54, in main
    suite = pyperf.BenchmarkSuite(benchmarks)  # filename='decimation.json')
  File "<REMOVED>/venv-local/lib/python3.8/site-packages/pyperf/_bench.py", line 639, in __init__
    self.add_benchmark(benchmark)
  File "<REMOVED>/venv-local/lib/python3.8/site-packages/pyperf/_bench.py", line 687, in add_benchmark
    name = benchmark.get_name()
[...]

PR #222, opened 2025-11-22
Add Runner.is_worker() helper method

amritamishra01:
This PR adds a new helper method Runner.is_worker() as suggested by @vstinner in issue #166.

bench_func() is executed both in the main process and in worker processes.
Only the main process returns a Benchmark object, while worker processes correctly return None.
However, user code currently has no clean way to detect whether it is running in a worker process, which can lead to unexpected None values ending up in lists.

This PR introduces:

✔ Runner.is_worker()

A small utility method that returns True if the current process was started with --worker, and False otherwise.

It simply exposes the existing worker detection logic via a public API, without changing any existing behavior.

This keeps the PR minimal exactly as requested.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -1,7 +1,9 @@
-.. _api:
-
 API
 ===
+
+.. module:: pyperf
+
+The module version...
 
 The module version can be read from ``pyperf.VERSION`` (tuple of int) or
 ``pyperf.__version__`` (str).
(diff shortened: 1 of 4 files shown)


doc/api.rst
```diff
@@ -1,8 +1,10 @@
-.. _api:
-
 API
 ===
 
+.. module:: pyperf
+
+The module version...
```
vstinner: What is this sentence which is not finished? I suggest removing it.

doc/api.rst
```diff
@@ -193,10 +195,12 @@ Benchmark class
 
       Format values including the unit.
 
-   .. method:: get_dates() -> (datetime.datetime, datetime.datetime) or None
+   .. method:: get_dates()
```
vstinner: Change unrelated to is_worker(), please revert it.

vstinner (commented):
Please revert unrelated changes.

vstinner (commented):
The PR title is "Add Runner.is_worker() helper method" but your PR doesn't add the method anymore, but is full of unrelated changes :-(

vstinner:
My review has not been addressed since last December. I close the PR. If someone wants to contribute a is_worker() method, please write a PR which only does that.

PR #224, opened 2026-01-02
`ruff check` and `ruff format` github action

maurycy:
This is very similar to what was done in [pyperformance](https://github.com/python/pyperformance):

- https://github.com/python/pyperformance/blob/main/.pre-commit-config.yaml
- https://github.com/python/pyperformance/blob/main/.github/workflows/lint.yml

`ruff check` is used since #184, and it's clean:

```
2026-01-23T20:50:23.718836000+0100 maurycy@gimel /Users/maurycy/work/pyperf (maurycy-ruff-ci) % uvx ruff check
All checks passed!
```

The PR also adds pre-commit via [prek](https://github.com/j178/prek).


vstinner:
As I wrote previously, I'm not convinced that it makes the code way easier to review. I don't think that this change is worth it. So I close this PR.

PR #230, opened 2026-01-24
Tachyon hook

maurycy:
The PR adds support for profiling benchmarks with a [Tachyon](https://docs.python.org/3.15/library/profiling.sampling.html) hook.

It's enabled only during benchmark execution. It supports nearly all flags as environment variables, and there's a guard for versions below 3.15. It's disabled on Windows, since I don't have access to any.

I'm not sure if this should be merged before October 2026, but it's a good way to verify Tachyon itself and offers a new way to profile the whole pyperformance suite.

Basic way to test:

```
uv run --python 3.15 -m pyperf timeit --hook tachyon '[x**2 for x in range(1000)]'
```

## pyperformance

I was able to run the whole `pyperformance` suite with it:

```
PYPERF_TACHYON_OPTS="--flamegraph" \
PYTHONPATH=/home/maurycy/github.com/maurycy/pyperf \
uv run --python 3.15 --with /home/maurycy/github.com/maurycy/pyperf/pyperformance \
python -c "import pyperformance.cli; pyperformance.cli.main()" run \
  --fast \
  --hook tachyon \
  --inherit-environ PYPERF_TACHYON_OPTS
```

## notes

- Beyond the scope, but the hook subsystem could be more granular (any interest?), so `WorkerTask` etc. wouldn't be included,
- `profiling.sampling` does not seem to offer a way to stop, except killing the process ([not the urgent priority](https://discuss.python.org/t/allow-use-of-tachyon-as-imported-python-module/105372/5)),
- What would be the best way to write a test for it?

### DIFF
--- a/pyperf/_hooks.py
+++ b/pyperf/_hooks.py
@@ -175,16 +175,16 @@ class perf_record(HookBase):
 class tachyon(HookBase):
     """Profile the benchmark using sampling profiler (Tachyon).
 
-    The value of the `PYPERF_TACHYON_EXTRA_OPTS` environment variable is
+    The value of the `PYPERF_TACHYON_OPTS` environment variable is
     appended to the `profiling.sampling attach` command line.
 
     This hook does not generate output filenames. Use -o or --output in
-    `PYPERF_TACHYON_EXTRA_OPTS` to control output. For most formats, -o can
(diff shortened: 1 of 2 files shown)


pablogsal:
Thanks for the PR @maurycy

PR #234, opened 2026-02-20
Add `--in-process` mode support

### LINKED ISSUE #233: Add `--in-process` runner mode for environments without subprocess support
## Problem Statement

pyperf currently requires subprocess spawning (`subprocess.Popen`) for every benchmark run. This makes pyperf unusable in environments where subprocess is not available.

The primary motivating environment is WebAssembly. WebAssembly platform has no subprocess support due to the fundamental platform level limitations of the WebAssembly sandbox. We ([Pyodide](https://github.com/pyodide/pyodide) maintainers) have an interest in running pyperformance benchmarks in WebAssembly to track the Python performance in the browser over time, and also benchmark the recent advancements such as tail call interpreter (https://github.com/pyodide/pyodide/issues/6102).

## Proposed Solution
Add a new `--in-process` flag in the runner that runs the entire benchmark in the current process.

## Implementation Plans

I think subclassing the `Manager` with an `InProcessManager` that overrides only [spawn_worker()](https://github.com/psf/pyperf/blob/3f0cfc828cbdda0f29210df93aacb02f798f761b/pyperf/_manager.py#L86) to execute WorkerProcessTask.create_run() directly instead of spawning a child process would work

## Caveats 
In-process mode will not provide the isolation benefits of subprocess:
- No clean memory state per run
- No OS-level CPU affinity control
For environments like Pyodide, this tradeoff is acceptable — subprocess is not an option at all. Results may have slightly different characteristics, but the benchmark values themselves are produced by the exact same timing code path.

## Implementation

I am willing to implement this and submit a PR. Feedback on the approach is welcome before I proceed.

ryanking13:
Resolve #233

Adds `--in-process` mode as proposed in #233. It enables running pyperf in browser-based environments like Pyodide.


vstinner (commented):
I would prefer to limit this PR to the `--in-process` option, and move the PyOdide changes to a separated PR (later).

ryanking13:
@vstinner Sure, sounds good. I remove pyodide-related changes from the PR.

ryanking13:
@vstinner @corona10 May I ask your review again when you have a chance?

PR #237, opened 2026-04-24
Hash pin GitHub Actions and add cooldown to Dependabot

hugovk:
Yet another compromise via unpinned GitHub Actions: https://socket.dev/blog/bitwarden-cli-compromised

Let's hash-pin GHA for our PyPI packages.

Done via `uvx gha-update`.

Also added 14-day cooldown (similar to cherry_picker) to Dependabot:

https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

PR #238, opened 2026-04-24
Add zizmor security linter

hugovk:
And fix its findings. Also add some other useful linters as well.

--

This will fail until https://github.com/psf/pyperf/pull/237 is merged, so in the meantime is also a good demo :)


.github/workflows/publish.yml
```diff
@@ -14,12 +19,12 @@ jobs:
...
+        persist-credentials: false
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
         python-version: '3.x'
-        cache: pip
-        cache-dependency-path: pyproject.toml
     - name: Install dependencies
       run: |
         python -m pip install --user --upgrade build
```
sethmlarson: Can we pin `build` to a version and hash, preferably a format that Dependabot accepts? (`.github/workflows/publish.txt` with version/hashes)

hugovk: Yep, see the newest commit.

vstinner:
> This will fail until https://github.com/psf/pyperf/pull/237 is merged, so in the meantime is also a good demo :)

I merged https://github.com/psf/pyperf/pull/237.

hugovk:
Updated!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.x"
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      - uses: j178/prek-action@cbc2f23eb5539cf20d82d1aabd0d0ecbcc56f4e3 # v2.0.2
+      - run: uvx safety check
```
vstinner: This command logs a deprecation message:

> DEPRECATED: this command (`check`) has been DEPRECATED, and will be unsupported beyond 01 June 2024.
>
> We highly encourage switching to the new `scan` command which is easier to use, more powerful, and can be set up to mimic the deprecated command if required.

When I run `uvx safety scan` locally, it asks me to create an account.

vstinner (commented):
This PR makes different changes not directly related to zizmor:

* Add a Lint job to GHA.
* Add a "safety check" step to the Lint job.
* Add `.github/workflows/publish.in` and `.github/workflows/publish.txt` used by `.github/workflows/publish.yml`. And update Dependabot rules to include `/.github/workflows`.
* Remove some empty lines at the end of two documentation files.
* Add a pre-commit configuration file (used by prek).

I think that I'm mostly fine with these changes, except of "safety check" which sounds redundant with Dependabot service. Is it really needed by zizmor?

.github/workflows/lint.yml
hugovk: We have this in cherry_picker and tbh I don't really like it, so let's remove it :)


```suggestion
```

hugovk:
> I think that I'm mostly fine with these changes, except of "safety check" which sounds redundant with Dependabot service. Is it really needed by zizmor?

It's not needed, I've removed it.

.github/workflows/publish.txt
```diff
@@ -0,0 +1,18 @@
+#
+# This file is autogenerated by pip-compile with Python 3.14
+# by the following command:
+#
+#    pip-compile --generate-hashes --output-file=/private/tmp/pyperf/.github/workflows/publish.txt /private/tmp/pyperf/.github/workflows/publish.in
```
vstinner: When and how is one supposed to update this file? Should it be updated when preparing a release? Is it regenerated by Dependabot?

Maybe add instructions to pyproject.toml header (the long comment)?

hugovk: > When and how is one supposed to update this file? ... Is it regenerated by Dependabot?

Yes, this PR adds the `package-ecosystem: pip` section to `.github/dependabot.yml`, similar to what we have in release-tools:

https://github.com/python/release-tools/blob/787921ddb0bb1a0d511795c5f7c58a79240489cc/.github/dependabot.yml#L14-L25

And then we get Dependabot PRs like https://github.com/python/release-tools/pull/364.

It'll open a PR once a month if needed. The 7 day cooldown means it won't consider an update within that time, which also gives supply-chain attack protection.

However, the schedule+cooldown don't apply to security updates -- Dependabot will open PRs for those sooner.

> Should it be updated when preparing a release? 

No need, it can be updated during by Dependabot as and when needed.
 
> Maybe add instructions to pyproject.toml header (the long comment)?

Yep, will do.

sethmlarson (approved):
Woohoo!

.github/workflows/publish.txt
vstinner: I would prefer to use a relative path. `/private/tmp/pyperf` doesn't exist on my system.

```suggestion
#    pip-compile --generate-hashes --output-file=.github/workflows/publish.txt .github/workflows/publish.in
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,22 @@
...
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.x"
```
maurycy, line 20: @hugovk 

What do you think about pinning it?

hugovk: I don't think it's necessary for the linter. GitHub likely won't bump it to 3.15 until some time after its final release, so everything should be ready by then.

But not a strong opinion against pinning.

vstinner (approved):
LGTM

vstinner:
Merged, thanks.

PR #246, opened 2026-07-14
Create SECURITY.md

De-ASI-INTERFACE:


vstinner (changes requested):
I don't understand this PR, I close it.

The documentation mentions version 4.0 and 5.1, whereas the latest pyperf release is 2.10.

"Reporting a Vulnerability" section seems to be copy/pasted without adapting the text pyperf.

PR #250, opened 2026-08-03
Move Ruff to pre-commit to pin it

hugovk:
`tox -e pep8` is failing in https://github.com/psf/pyperf/pull/248 and https://github.com/psf/pyperf/pull/249.

This is because they run `ruff check`, and Ruff has just released version 0.16 which increases the number of default rules from 59 to 413: https://astral.sh/blog/ruff-v0.16.0

This PR moves the `ruff check` from tox to pre-commit, thereby pinning it, so we won't get any surprises outside explicit updates of this pin.

I pinned it to the previous 0.15, because there will be a lot of changes to go to 0.16, and we can handle that in another PR, including deciding whether to ignore some rules.


vstinner (approved):
LGTM

vstinner:
Merged, thanks for the fix.

PR #251, opened 2026-08-03
Update Python versions

hugovk:
Add:
* Python 3.13-3.15
* Python 3.14-3.15 free-threaded
* PyPy 3.11

Drop:
* Python 3.9: EOL
* PyPy 3.9-3.10: EOL
* Python 3.13 free-threaded: experimental
* Details about which version to use for Python 2: no longer important

Change:
* Test free-threaded via GitHub Actions rather than deadsnakes. I think GHA's free-threading wasn't yet supported when we initially added it.
* Update PyPI links.


vstinner (approved):
LGTM

vstinner:
pre-commit fails with: `Error: pyperf/_hooks.py:21:8: UP036 Version block is outdated for minimum Python version`

Code:

```py
def get_hooks():
    hook_prefix = "pyperf.hook"
    entry_points = importlib.metadata.entry_points()
    if sys.version_info[:2] < (3, 10):  # <======= HERE =====
        group = entry_points[hook_prefix]
    else:
        group = entry_points.select(group=hook_prefix)
    return group
```

The code path for Python 3.9 and older can now be removed.

vstinner:
Merged, thanks.

PR #254, opened 2026-08-27
refactor: Made _T_DIST_95_CONF_LEVELS a tuple because it's a constant…

DebanKsahu:
# Description #
- I was reading the code and saw that `_T_DIST_95_CONF_LEVELS` never modified in the later part of `_utils.py` file as it is only accessed so I thought it is reasonable to make it a tuple instead of the dynamic list.
- This has no significant performance gain, but I think it would avoid the `amortized space allocation` for **dynamic list** in python which happens to avoid `amortized O(N)` complexity for `append` operation


vstinner (approved):
This is really a pedantic change... It's a private constant in a private module, the constant is only used by `tdist95conf_level()`.

> This has no significant performance gain, but I think it would avoid the amortized space allocation for dynamic list in python which happens to avoid amortized O(N) complexity for append operation

`_T_DIST_95_CONF_LEVELS` is never modified, so this point is irrelevant.

I only approved the change to avoid someone else proposing a similar change in the future.

DebanKsahu:
Sorry, I will be careful with the future code changes if I made any.

PR #257, opened 2026-09-05
Improving type system for pyperf library

### LINKED ISSUE #255: Improve Type System for `pyperf`
# Description #
- I recently came across `pyperf` because I want to do some benchmark testing of my changes which I did on an opensource project named `ArchUnitPython` which are generally to improve the performance of the library. Initially I wanted to start with inbuilt `time` module but later I found `pyperf` which kind of can produce good benchmark by handling things like cold start, number of process spawn, warm benchmark etc. But during coding I found that the type system of `pyperf` is not good as most of most return type and function parameter have `Unknown` type which kind of hindering my auto completion and understanding of that function or parameter.
- Eg;

Image

Image

- Now it might be issue from `pyrefly` typechecker but the thing is the library on which I contributed and benchmarking my changes have no `Unknown` **(Till now)** same for other libraries like `FastAPI`, `PydanticAI` and some more.
- Because of this I am planning to do contribution to the library and improve the type system of `pyperf`. 

If any maintainer could guide me or discuses in this issue, then it would be helpful.

DebanKsahu:
# Description #
## _utils.py ##
- In `tdist95conf_level` function the doc string clearly mention that `df` means degree of freedom and expected to be a `int` that's why I type annotated it with `int`. Now that `tdist95conf_level` showing that it required a `int degree of freedom so there no need for `df = int(round(df))` as it's now user's reponsibility to pass correct value.
- `pooled_sample_variance`, `tscore` and `is_significant` all required input of type `Sequence[float]` because the inputs are iterated over many time so we can't accept an `Iterable[float]` because it will exhaust the base list.

Fixes #255


vstinner:
I understood that having wrong type annotations can be worse than not having type annotations. We should check if type annotations are correct using a tool (such as mypy) to make sure that they remain correct when the code will evolve. What is your plan to check type annotations?

Does this PR add all type annotations? Or do you plan further changes?

DebanKsahu:
I am planning to add all type annotations in this PR only. For the testing part I am planning to use `pyrefly`, we can add the settings to `pyproject.toml` file only and add this test into the commit hook just like `ruff`.

vstinner:
Would it be possible to include the type check integration in this change? Otherwise, I don't know how to test this change locally. How can I test that the added type annotations are correct?

DebanKsahu:
Yeah, I was planning to add type checking test at the end but, as you said I will add those test in the next commit. Should I go with Pyrefly?

vstinner:
> Should I go with Pyrefly?

As you want, I don't have any preference.

sergey-miryanov:
I suggest using two or more type checkers (e.g., `mypy` and `pyrefly`). In my experience, I've encountered cases where `pyrefly` had a bug and missed a type error, while `mypy` caught it.

DebanKsahu:
> I suggest using two or more type checkers (e.g., `mypy` and `pyrefly`). In my experience, I've encountered cases where `pyrefly` had a bug and missed a type error, while `mypy` caught it.

Sure then, I will go with `pyrefly` and `mypy`.

vstinner:
Aha, I see that you added configuration to pyproject.toml.

So what's the way to run these tests? Just run "pyrefly" and "mypy"?

It may be good run these tools in the GitHub Action. Tell me if you need help with that.

Also, your PR is still marked as a draft. It's not ready for a review?

DebanKsahu:
- You can use `pyrefly check` for pyrefly and `mypy` for mypy. If you are using uv already installed these tools globally then you can also try `uv tool run pyrefly check` and `uv tool run mypy` respectively. 
- Yeah adding them into github action seems good but before that we should finalize which rules to stay and which are not.
- Currently there are other files which are yet to get type annotation so that's why this pr is still in draft, currently my exams are going on so work is progressing slowly, once all files get annotated I will make this PR open.

PR #260, opened 2026-09-13
Allow zero-valued memory metadata

### LINKED ISSUE #259: pyperf fails when `getrusage()` returns zero for `ru_maxrss`
Pyperf fails with `ValueError: invalid metadata 'mem_max_rss' value: 0` because `resource.getrusage()` returns zero for `ru_maxrss`.
This is more likely to happen on small benchmarks with minimal memory footprint (e.g. pyperformance::raytrace)

`getrusage()` [(source)](https://codebrowser.dev/linux/linux/kernel/sys.c.html#getrusage) calculates `ru_maxrss` based on `get_mm_rss()` [(source)](https://codebrowser.dev/linux/linux/include/linux/mm.h.html#get_mm_rss):
```
static inline unsigned long get_mm_rss(struct mm_struct *mm)
{
    return get_mm_counter(mm, MM_FILEPAGES) +
        get_mm_counter(mm, MM_ANONPAGES) +
        get_mm_counter(mm, MM_SHMEMPAGES);
}
static inline unsigned long get_mm_counter(struct mm_struct *mm, int member)
{
    return percpu_counter_read_positive(&mm->rss_stat[member]);
}
```
`percpu_counter_read_positive()` can return zero [(source)](https://codebrowser.dev/linux/linux/include/linux/percpu_counter.h.html#percpu_counter_read_positive):
```
/*
 * It is possible for the percpu_counter_read() to return a small negative
 * number for some counter which should never be negative.
 *
 */
static inline s64 percpu_counter_read_positive(struct percpu_counter *fbc)
{
    s64 ret = READ_ONCE(fbc->count);
    if (ret >= 0)
        return ret;
    return 0;
}
```
Those percpu_counters are approximate as it stated in the very beginning [(source)](https://codebrowser.dev/linux/linux/include/linux/percpu_counter.h.html):
```
/*
 * A simple "approximate counter" for use in ext2 and ext3 superblocks.
 */
```
It's completely legal for `getrusage()` to return zero for `ru_maxrss`, this happens on some hosts, and pyperf fails in such scenarios.

Allowing zeroes for BYTES in `pyperf/_metadata.py` solves the issue:
```
BYTES = _MetadataInfo(format_filesize, (int,), is_strictly_positive, 'byte')
```
->
```
BYTES = _MetadataInfo(format_filesize, (int,), is_positive, 'byte')
```

GruffElixir:
## Summary

Allow zero for memory usage metadata because `resource.getrusage()` can legally report `ru_maxrss` as zero on Linux. Other byte metadata keeps the same validation, and loop counts remain strictly positive.

Fixes #259

## Test Plan

- `python -bb -Wd -m unittest discover -s pyperf/tests/ -q`


vstinner (commented):
LGTM

vstinner (approved):

